PR #58, opened 2016-12-26
Multi-sequence support for all MultiTransformBlocks

MilesCranmer:
- Adds multiple sequence support to all MultiTransformBlocks with the self.trigger_sequence switch. 
    - In the read/write loop, set `self.trigger_sequence=True` and modify the headers and gulp_sizes appropriately, and the next loop iteration will allocate outspans on new sequences (it triggers new sequences on multiple output rings).
- NumpyBlock and NumpySourceBlock create new sequences automatically when their input headers or outputted data changes.
- Also creates `_main()` function in MultiTransformBlocks, which performs tedious activities like setting affinity.core and creating empty headers for the rings. The `_main()` function is called automatically by the pipeline and calls `main()` when finished.


python/bifrost/block.py
```diff
@@ -275,37 +284,44 @@ def read(self, *args):
...
+                with nested(*[out_ring.begin_sequence(
+                    str(int(time.time()*1000)),
+                    int(time.time()*1000),
+                    header=json.dumps(self.header[ring_name]),
+                    nringlet=1) \
+                        for out_ring, ring_name in self.izip(out_rings, args)]) as out_sequences:
+
+                    self.trigger_sequence = False
+                    #TODO: Eventually this could be used on each ring individually.
+                    while not self.trigger_sequence:
```
benbarsdell: I suggest adding the text from the PR description somewhere here; otherwise what trigger_sequence is and how it works is a bit opaque.

MilesCranmer: Good point. I will add some documentation to that variable. 

Thanks for the review!
Cheers,

Miles

MilesCranmer: I have added it. Let me know if you think the rest looks good.

Cheers,
Miles

benbarsdell:
Is this test failure a problem?

MilesCranmer:
It might be another problem with the CPU-only build of Bifrost. My GPU build of this branch works for all 86 tests. Have you seen a problem on your machine?

benbarsdell:
You need to merge or rebase onto master locally before testing (e.g., `git checkout feature/multi-sequence-support ; git rebase master`). I get lots of test errors after doing this.

MilesCranmer:
Strange—I am merging master before building. The only conflict was whitespace inside of `config.mk`.
I can see all of your new unit tests running as well.

What errors are you seeing?

benbarsdell:
Oops... sorry, my dumb mistake; I forgot to "setup.py install" it before running the tests :|. I now see no failures.

I guess the jenkins failure is just the same thing we've been seeing in master? (Initially I had thought it was something new.)

MilesCranmer:
Does this look good for merging? I should be pushing the travis-ci fix later today.

PR #74, opened 2017-02-10
Testbench and some reorganization

telegraphic:
Hey all,

Here's a few testbenches that I think would work well as tutorials too. There's a few things going on here (sorry):

* I created a directory called 'blocks', in which I propose we move all blocks. At the moment I've added relative blocks via imports in the blocks/__init__.py script

* I've created a sink and source block for reading/writing binary data using numpy's fromfile() and tofile() commands. 

* I made a very basic FFT block that wraps the bfCUDA FFT. 

* I added a testbench directory, which can be used for running basic end-to-end pipeline tests and is also good as self-documentation.

* I wrote a testbench for the binary file read + write blocks. This works well.

* I wrote a testbench for the FFT, using which I get odd results with the testbench that indicate something is broken. I wonder if there's a device synchronize problem in the fft.cu code?

* I also added four very basic cuda kernels for basic operations (+ -  / * ). There is no doubt a cooler way to do this. Adding this is all part of my goal of making bifrost a builder's toolkit a la CASPER.


python/bifrost/fft_block.py
```diff
@@ -0,0 +1,46 @@
...
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from pipeline import TransformBlock
+from copy import deepcopy
+from fft import fft
+
+class FFTBlock(TransformBlock):
```
MilesCranmer: I wonder if we should have this block handle both CPU and GPU FFT's, or maybe rename one of them so there is no conflict.
Thoughts?

MilesCranmer:
Great work Danny, this looks good to me! I will test the code out now.

python/bifrost/fft_block.py
benbarsdell: Ideally it should look at what space the inputs and outputs are to decide between CPU and GPU.

I've been working on a new FFT block that will add some features, but it's proving a bit more complicated than I thought.

telegraphic: Agreed, that would be nice. Can CPU/GPU select be done once at instantiation?

No strong opinion on if we should merge examples and testbench. Maybe testbench and tutorials would be a better split?

MilesCranmer:
This looks good for merging!

Maybe after, we could follow your idea and reorganize the block directories? We'll also need to rewrite some unit tests (and other files, e.g., example pipelines) to import correctly.

MilesCranmer:
(I just realized that Ben created a commit which did this already)

docs/source/your-first-blocks.rst
```diff
@@ -0,0 +1,286 @@
...
+            
+        def on_data(self, ispan, ospan):
+            in_nframe  = ispan.nframe
+            out_nframe = in_nframe
+            
+            idata = ispan.data 
+            odata = ospan.data
+            
+            odata[...] = idata + self.n_to_add
+            return out_nframe
```
benbarsdell: You don't actually need the return statement if you processed the whole span :)

telegraphic:
Now updated with more documentation

MilesCranmer:
Looks good to me. Merging.

MilesCranmer:
By the way, you could also tell travis to execute the new unit tests you added if they are all non-GPU.

PR #77, opened 2017-03-03
LWA-SV developments

jaycedowell:
A couple of things I've added to help with ADP development at LWA-SV:
* bfUdpTransmit class for sending out generic UDP data packets, 
* logging for bfUdpCapture/Transmit to /dev/shm/bifrost/ for monitoring network operations, and
* CI4/I4 packing in quantizer.cpp.

### DIFF
--- a/python/bifrost/udp_transmit.py
+++ b/python/bifrost/udp_transmit.py
@@ -1,20 +1,20 @@
 # -*- coding: utf-8 -*-
-from libbifrost import _bf, _check, _get, _string2space, _space2string
+from libbifrost import _bf, _check, _get
 
 import ctypes
 import numpy as np
 
 def _packet2pointer(packet):
-	buf = ctypes.create_string_buffer(packet)
-	len = ctype.sizeof(buf)
-	return ctypes.pointer(buf), len
+	buf = ctypes.c_char_p(packet)
+	siz = ctypes.c_uint( len(packet) )
+	return buf, siz
 
 
 def _packets2pointer(packets):
-	count = len(packets)
-	buf = ctypes.create_string_buffer("".join(packets))
-	len = ctypes.sizeof(buf)/count
-	return ctypes.pointer(buf), len, count
+	count = ctypes.c_uint( len(packets) )
+	buf = ctypes.c_char_p("".join(packets))
+	siz = ctypes.c_uint( len(packets[0]) )
+	return buf, siz, count
 
 
 class UDPTransmit(object):
@@ -30,9 +30,9 @@ class UDPTransmit(object):
 	def __exit__(self, type, value, tb):
 		pass
 	def send(self, packet):
-		ptr, len = _packet2pointer(packet)
-		return _get( _bf.UdpTransmitSend(self.obj, ptr, len) )
+		ptr, siz = _packet2pointer(packet)
+		return _get( _bf.UdpTransmitSend(self.obj, ptr, siz) )
 	def sendmany(self, packets):
 		assert(type(packets) is list)
-		ptr, len, count = _packets2pointer(packets)
-		return _get( _bf.UdpTransmitSendMany(self.obj, ptr, len, count) )
+		ptr, siz, count = _packets2pointer(packets)
+		return _get( _bf.UdpTransmitSendMany(self.obj, ptr, siz, count) )
--- a/python/bifrost/udp_transmit.py
+++ b/python/bifrost/udp_transmit.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+from libbifrost import _bf, _check, _get, _string2space, _space2string
+
+import ctypes
+import numpy as np
+
+def _packet2pointer(packet):
+	buf = ctypes.create_string_buffer(packet)
+	len = ctype.sizeof(buf)
+	return ctypes.pointer(buf), len
+
+
+def _packets2pointer(packets):
+	count = len(packets)
+	buf = ctypes.create_string_buffer("".join(packets))
+	len = ctypes.sizeof(buf)/count
+	return ctypes.pointer(buf), len, count
+
+
+class UDPTransmit(object):
+	def __init__(self, sock, core=-1):
+		self.obj = None
+		self.obj = _get(_bf.UdpTransmitCreate(fd=sock.fileno(),
+		                                     core=core), retarg=0)
+	def __del__(self):
+		if hasattr(self, 'obj') and bool(self.obj):
+			_bf.UdpTransmitDestroy(self.obj)
+	def __enter__(self):
+		return self
+	def __exit__(self, type, value, tb):
+		pass
+	def send(self, packet):
+		ptr, len = _packet2pointer(packet)
+		return _get( _bf.UdpTransmitSend(self.obj, ptr, len) )
+	def sendmany(self, packets):
+		assert(type(packets) is list)
+		ptr, len, count = _packets2pointer(packets)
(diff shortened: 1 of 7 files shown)


src/udp_capture.cpp
```diff
@@ -595,6 +605,56 @@ class BFudpcapture_impl {
...
+		std::string log_dir = "/dev/shm/bifrost/"+std::to_string(this->_pid);
+		int status = system(("mkdir -p "+log_dir).c_str());
+		if( status != 0 ) {
+			// TODO: What to do here?
+			throw std::runtime_error("Cannot create capture status directory");
+		}
+		
+		// Open the log files
+		//// Capture type
+		_type_log.open((log_dir+"/type").c_str(), std::fstream::out);
```
benbarsdell: Can remove `std::fstream::out` if these are `std::ofstream`.

src/udp_capture.cpp
```diff
@@ -595,6 +605,56 @@ class BFudpcapture_impl {
 		size_t total_span   = contig_span * 4;
 		size_t nringlet_max = 1;
 		_ring.resize(contig_span, total_span, nringlet_max);
+		this->open_logs();
+	}
+	inline void open_logs() {
```
benbarsdell: Can make this private since it's only called by the constructor.

src/udp_capture.cpp
```diff
@@ -631,12 +691,14 @@ class BFudpcapture_impl {
...
-		     << stats->ninvalid << ", "
-		     << stats->nlate << endl;
+		_stat_log.seekp(0);
+		_stat_log << "ngood_bytes, nmissing_bytes, nvalid, ninvalid, nlate = "
+		           << _ngood_bytes << ", "
+		           << _nmissing_bytes << ", "
+		           << stats->nvalid << ", "
+		           << stats->ninvalid << ", "
+		           << stats->nlate << endl;
+		_stat_log.flush();
```
benbarsdell: `endl` will implicitly flush, so this shouldn't be needed.

src/udp_capture.cpp
```diff
@@ -658,20 +720,27 @@ class BFudpcapture_impl {
...
 				_payload_size = pkt->payload_size;
-				cout << "CHAN0 " << _chan0 << endl;
-				cout << "NCHAN " << _nchan << endl;
-				cout << "PAYLOAD_SIZE " << pkt->payload_size << endl;
+				_chan_log.seekp(0);
+				_chan_log << "CHAN0, NCHAN, PAYLOAD_SIZE = " 
+				           << _chan0 << ", " 
+				           << _nchan << ", "
+						 << _payload_size << endl;
+				_chan_log.flush();
```
benbarsdell: Same as above re `endl` => `flush`.

src/udp_capture.cpp
```diff
@@ -718,6 +787,7 @@ BFstatus bfUdpCaptureCreate(BFudpcapture* obj,
 }
 BFstatus bfUdpCaptureDestroy(BFudpcapture obj) {
 	BF_ASSERT(obj, BF_STATUS_INVALID_HANDLE);
+	obj->close_logs();
```
benbarsdell: This could be called from the destructor. Even better would be to wrap the `mkdir/rm` in a separate class's constructor and destructor, which would remove the need for `close_logs()` completely.

src/udp_capture.cpp
```diff
@@ -595,6 +605,56 @@ class BFudpcapture_impl {
...
+		_type_log.close();
+		_size_log.close();
+		_chan_log.close();
+		_stat_log.close();
+		
+		// Cleanup the process directory
+		int status = system(("rm -rf /dev/shm/bifrost/"+std::to_string(this->_pid)).c_str());
+		if( status != 0 ) {
+			// TODO: What to do here?
+			throw std::runtime_error("Cannot remove capture status directory");
```
benbarsdell: I'm not sure exactly what to do here either, but probably best to avoid throwing because this is essentially a destructor.

src/bifrost/udp_transmit.h
```diff
@@ -0,0 +1,27 @@
+#ifndef BF_UDP_TRANSMIT_H_INCLUDE_GUARD_
```
benbarsdell: License header

python/bifrost/udp_transmit.py
```diff
@@ -0,0 +1,38 @@
...
+		self.obj = None
+		self.obj = _get(_bf.UdpTransmitCreate(fd=sock.fileno(),
+		                                     core=core), retarg=0)
+	def __del__(self):
+		if hasattr(self, 'obj') and bool(self.obj):
+			_bf.UdpTransmitDestroy(self.obj)
+	def __enter__(self):
+		return self
+	def __exit__(self, type, value, tb):
+		pass
```
benbarsdell: The best thing to do here might be to call a `bfUdpTransmitClose` method that closes the logs. `bfUdpTransmitDestroy` would then only call `close_logs` if it hadn't already been called.

python/bifrost/udp_transmit.py
```diff
@@ -0,0 +1,38 @@
...
+
+def _packet2pointer(packet):
+	buf = ctypes.c_char_p(packet)
+	siz = ctypes.c_uint( len(packet) )
+	return buf, siz
+
+
+def _packets2pointer(packets):
+	count = ctypes.c_uint( len(packets) )
+	buf = ctypes.c_char_p("".join(packets))
```
benbarsdell: One way to avoid overhead here would be to have `bfUdpTransmitSendMany` accept one array for the headers and a separate one for the payloads, and to use the gather capabilities of `sendmmsg` to stitch them together implicitly. This would allow passing a block of data as-is (with the headers generated as a separate string).

src/udp_transmit.cpp
```diff
@@ -0,0 +1,257 @@
...
+		
+		return BF_TRANSMIT_CONTINUED;
+	}
+	BFudptransmit_status sendmany(char *packets, unsigned int len, unsigned int npackets) {
+		ssize_t state;
+		unsigned int i;
+		struct mmsghdr *mmsg = NULL;
+		struct iovec *iovs = NULL;
+		
+		mmsg = (struct mmsghdr *) malloc(sizeof(struct mmsghdr)*npackets);
```
benbarsdell: Could use `std::vector` and `resize` here to reuse the memory allocation.

src/udp_transmit.cpp
```diff
@@ -0,0 +1,257 @@
+#include "assert.hpp"
```
benbarsdell: License header

src/udp_capture.cpp
jaycedowell: I also think there is some room for improvement in the logging since the UDP transmit stuff ends up in a separate /dev/shm/bifrost_transmit directory.  It would be cleaned if they could all be in the same /dev/shm/biforst/<PID> directory.  However, that does require being more careful about cleaning up at the end.

python/bifrost/udp_transmit.py
jaycedowell: I'm inclined to keep the logs open until the object is destroyed since it matches the current behavior for bfUdpCapture.

python/bifrost/udp_transmit.py
jaycedowell: For a first pass at this I wanted something that was very generic and did not make too many assumptions about the packets (this only assumes that all of the packets are the same size).  In the future it would be nice to have something similar to what you suggested but with the classes creating the headers on-the-fly depending on the packet format.

PR #84, opened 2017-05-24
Some bugfixes in merge_axis and fdmt update

telegraphic:
* `views.merge_axis`: Minor bugfix in merge_axes (units were not being merged).
* `views.merge_axis`: Added 'ignore_units' argument to merge_axes to allow force merges (Ben does not like this cavalier approach)
* `views.merge_axis`: Learned the magic of vim's :retab command to convert tabs to spaces.
* `blocks.print_header`: Changed print_header block to use pprint
* `blocks.guppi_raw`: Added 'gulp_nframe' to guppi_raw tensor to allow for split_axes views to work.
* `testbench/test_fdmt.py`: Updated test_fdmt.py testbench to get it working with newer labels (eg 'dispersion' instead of 'dispersion measure')


benbarsdell:
This is great, thanks Danny.

One thing: changing tabs->spaces along with actual code changes in a single commit makes it impossible to see what was changed. We should really separate them into different commits.

Btw you can avoid unwanted merge commits by using "git pull --rebase" to rebase your branch on top of master instead of merging.

telegraphic:
I totally agree about the tabs vs. spaces, and was sad to see the diff. It may anger you to know I've been using vim not emacs too.

Just writing up some notes that will become documentation about the bfmap function, and noted some examples in the docstirng were broken as the shape=None argument was making shape='i' in the call.

python/bifrost/units.py
```diff
@@ -31,13 +31,16 @@
...
+    try:
+        new_quantity = old_quantity.to(new_units)
+    except pint.DimensionalityError:
+        raise ValueError("Cannot convert units %s to %s" %
 		                 (old_units, new_units))
-	return new_quantity.magnitude
+    if value >= 0:
+        return new_quantity.magnitude
+    else:
+        return -1 * new_quantity.magnitude
```
benbarsdell: Negative values seem to work fine for me already; can you confirm that this is necessary?

telegraphic: Hmm, well, I only added it because my pipeline was crashing... I guess it depends on whether the `pint` library is returning the mathematical definition of magnitude, or something else (which actually seems more likely what the authors intended!)

benbarsdell: OK. If the problem is in merging time axes from a guppi raw block, you may need to use a ReverseBlock.

telegraphic: That may indeed be where it comes from. Here's an example pipeline to show where it's popping up:
```python
with bfp.Pipeline() as pipeline:
    
    n_blocks = 128              # Number of blocks in raw file
    n_coarse = 64
    n_t_per_coarse = 524288
    n_pol    = 4                
    n_int    = 128              # Runtime - Number of integrations
    n_chunks = 2                # Runtime - Number of time chunks 
    n_t      = n_blocks / n_int
    n_chan   = n_coarse * n_t_per_coarse * n_chunks
    
    filelist = sorted(glob.glob('*.raw'))[:2]
    
    # Read from guppi raw file
    b_guppi   = blocks.read_guppi_raw(filelist, core=1, buffer_nframe=4)
    b_gup2    = views.rename_axis(b_guppi, 'freq', 'channel')
    # Buffer up two blocks & reshape to allow longer FFT
    with bf.block_scope(fuse=True, core=2):
        b_gup2    = views.split_axis(b_gup2, axis='time', n=n_chunks, label='time_chunk')
        b_gup2    = blocks.transpose(b_gup2, axes=['time', 'channel', 'time_chunk', 'fine_time', 'pol'], buffer_nframe=1)
        b_gup2    = views.merge_axes(b_gup2, 'time_chunk', 'fine_time', label='fine_time')
    
    # Copy over to GPU and FFT
    with bf.block_scope(fuse=True, core=3):
        b_copy    = blocks.copy(b_gup2,  space='cuda')
        b_fft     = blocks.fft(b_copy,  axes='fine_time', axis_labels='freq')
        b_ffs     = blocks.fftshift(b_fft, axes='freq')
        b_pow     = blocks.detect(b_ffs, mode='stokes')

    # Copy back to host and flatten channel/freq axis to form output spectra
    b_copp    = blocks.copy(b_pow, space='system', buffer_nframe=1, core=4)
    #
    b_flat    = views.merge_axes(b_copp, 'channel', 'freq', label='freq')
    blocks.print_header(b_flat)
    pipeline.run()
```

And the exception:

```
Exception in thread Pipeline_0/PrintHeaderBlock_0:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "build/bdist.linux-x86_64/egg/bifrost/pipeline.py", line 316, in run
    self.main(active_orings)
  File "build/bdist.linux-x86_64/egg/bifrost/pipeline.py", line 453, in main
    self.sequence_proclogs[i].update(iseq.header)
  File "build/bdist.linux-x86_64/egg/bifrost/ring2.py", line 288, in header
    hdr = self.header_transform(deepcopy(hdr))
  File "build/bdist.linux-x86_64/egg/bifrost/views/basic_views.py", line 142, in header_transform
    "%f != %f" % (scale1, n*scale2))
ValueError: Scales of merge axes do not line up: -2.929688 != 2.929688
```

telegraphic: Not sure I quite understand the need for an axis reverse: the freq axis generally does run negative, but in this instance doesn't merging still result in monotonically increasing frequency deltas? 

I can see that certain combinations of axes orders could produce unwanted periodic flips, but a general formalism escapes me at the moment :/

telegraphic: (PS: I'm out of space on the GPU so reverse is tricky -- I can write a numpy block version easy enough if you convince me though!)

benbarsdell: My interpretation is that you need to reverse the fine_time axis before merging, and then probably again after re-splitting. There could be something else going on though. We could try plotting the data with/without the reversal(s).

The space problem is a bit tricky. At some point I want implement a fused fft+fftshift+detect, which will help.

telegraphic: To confirm, pint 'overloaded' the term 'magnitude' to mean 'value, not magnitude':
```python
In [1]: import pint

In [2]: ureg = pint.UnitRegistry()

In [3]: a = 1.0 * ureg.meters

In [4]: a
Out[4]: <Quantity(1.0, 'meter')>

In [5]: a.magnitude
Out[5]: 1.0

In [6]: a = -1.0 * ureg.meters

In [7]: a.magnitude
Out[7]: -1.0
```

MilesCranmer: It looks like it. I'll raise an issue asking them about it on their github: maybe they can compromise with a synonymous property called `value` (because I agree that magnitude seems a bit unintuitive in that case...).

telegraphic: It seems a little ironic that a library about SI units gets the definition of magnitude wrong :P

python/bifrost/blocks/guppi_raw.py
```diff
@@ -59,7 +59,7 @@ def on_sequence(self, reader, sourcename):
         df_MHz = bw_MHz / nchan
         f0_MHz = cfreq_MHz - 0.5*(nchan-1)*df_MHz
         # Note: This will be negative if OBSBW is negative, which is correct
-        dt_s   = 1. / df_MHz / 1e6
+        dt_s   = np.abs(1. / df_MHz / 1e6)
```
benbarsdell: With a reversed channel axis (-ive scale), the fine_time axis is also reversed (-ive scale). The FFT preserves the sign of the axis scale, so if you FFT+fftshift the fine_time axis, then the result will still have -ive scale, and so it can be merged with the channel axis (both have -ive scales).

This means the fine_time axis is reversed relative to the coarse time (block) axis, so merging fine_time and coarse time will require use of a ReverseBlock I think.

telegraphic: Hey Ben, this is one of those confusing things that doesn't really make sense. Perhaps more flexible compromise for the end user while being 'correct' is adding a `reverse_scale` view to flip the -ive sign? 

The `fine_time` axis *has* to run in the same direction as the `time` axis IMO. Negative time isn't really physical, it arises due to degeneracy between the negative frequency introduced by mixing with an LO and taking the upper sideband, i.e. `f_IF = f_LO - f_RF`. (do we agree on this?)

I agree that the FFT preserves the sign of the axis scale, and that `coarse channel` can be considered -ive scale.  I think the key is *can be considered*, and that if it is considered, you have to consider *both* `time` and `fine_time` to be -ive. But time definitely flows forward ;)

Another way of thinking about it that I haven't squared away: if we looked at a single pulse from a pulsar (squared each coarse channel and did an imshow), it would be flipped in frequency but not time?

benbarsdell: Interesting. Could the problem be that OBSBW being negative really means that OBSFREQ is a negative frequency? I.e., what if we moved the sign of OBSBW to OBSFREQ instead, so the axis scale has a negative starting frequency but a positive step. All axes would then have positive steps and merging becomes no problem.

It would be interesting to check if a negative starting frequency works in the FDMT code; I have a feeling it will. Actually, it looks like it may just require some minor numerical changes to avoid complex numbers appearing in the delay computation (the final result is always real).

telegraphic: Hmm. I think keeping OBSFREQ as-is for now is pragmatic. I'm happy with the ability to change step in my branch for now!

MilesCranmer:
What is the status on this PR, @benbarsdell and @telegraphic ? Could we merge it? I'd be sad to see it to start accumulating merge conflicts.

PR #96, opened 2017-06-26
Merge initial round of benchmarks

MilesCranmer:
- Create benchmarks modules (optional install inside test/benchmarks)
    - Adds `PipelineBenchmarker` class, which is basically an entire class built around @telegraphic's timeit function, which lets you record timeit's for various **specific** parts of your pipeline and its blocks
    - (Can probably do some of this with `/dev/shm` logging and Jayce's utility's)
- Log of historical benchmark scores in .csv. This will probably be machine-generated later by a CI service, but I have put an example in there right now
- Benchmark for compile time of Bifrost
- Benchmark for `read_sigproc` block
- WIP Benchmark for a GPU FFT pipeline
- Syntax benchmark comparing Danny's gpuspec code with Bifrost against the legacy, pure-C++ code
    - test/benchmarks/development_vs_gpuspec/Makefile
    - Run with `make test`, the output is:
```
With Bifrost:
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
       165       6.1     1.2       51.9       16            0      0.00    0.00
==========================================================================================
Without Bifrost:
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
      1329      23.0     5.2      247.8       51            1      0.02    0.41
==========================================================================================
```

I do some preprocessing on Danny's code to remove the extra (?) Hdf block, and only select files in both repo's that I think are relevant to the calculation. See WITH_BIFROST and WITHOUT_BIFROST for the list of files I include. The above numbers are important to get right as they are the only benchmarks scores (?) to be listed in the paper other than packet capture rate/other LWA Sev pipelines. This is pending further discussion, but that is what I recall we had decided.

Output via [lizard](https://github.com/terryyin/lizard)


MilesCranmer:
New output, taking into account Danny's comments during latest telecon (on which files to use). I also do zero preprocessing now, because it doesn't look like there's an extra Hdf* block.

```
With Bifrost:
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
       177       5.5     1.2       49.5        8            0      0.00    0.00
==========================================================================================
Without Bifrost:
==========================================================================================
Total nloc   Avg.NLOC  AvgCCN  Avg.token   Fun Cnt  Warning cnt   Fun Rt   nloc Rt
------------------------------------------------------------------------------------------
      1329      23.0     5.2      247.8       51            1      0.02    0.41
==========================================================================================
```

This is the list of files currently used:

- Bifrost (telegraphic/bunyip): blocks/accumulate.py blocks/hdf.py bf_gpuspec.py
- Non-Bifrost (UCBerkeleySETI/gbt_seti): guppi2spectra.c guppi2spectra_gpu.cu guppi2spectra_gpu.h filterbank.h

@benbarsdell, let me know if you think this is merge-able. The work is all inside a new folder: test/benchmarks, so there should not be any conflicts with the main Bifrost code

MilesCranmer:
This PR now also includes a variety of speed ups to the Bifrost python frontend, which were shown by the profiler to be major hotspots. Right now these mostly are cases where a PyCLibrary constant was being generated at each hit, rather than being cached by Python. The most expensive cases of this are now globals in their respective files.

MilesCranmer:
Latest commits introduce many more performance improvements, as well as benchmark numbers for the linear FFT pipeline courtesy of Ben.

The performance improvements include a commit be373ec which takes all pyclibrary calls in the main Bifrost API and replaces them with a global variable. The performance improvement from this and the few instances where error checking is put into __debug__ statements is quite significant. You need to install with `python setup.py install -O2`, and run with `python -OO ...` to see some of this speedup.

After benchmarking against skcuda for an FFT->IFFT pipeline using the GPU, here's the performance figure that's going in the paper:

screenshot 2017-07-20 19 25 57.

Also, if possible, @benbarsdell, we should merge this soon, and then up the version of Bifrost and make a release so we can reference the version number in the benchmarks described in the paper.

PR #98, opened 2017-07-31
Replace pyclibrary with ctypesgen

benbarsdell:


MilesCranmer:
Heroic effort, Ben!

I am testing this quickly on AWS and then will merge.

MilesCranmer:
Benchmark running on AWS, all timings look faster.

If the standard deviations look reasonably small, this can probably be included in the paper. The plots are all automatically generated so I can update them easily. We basically just need to run `s/\cpyclibrary/ctypesgen/g` on new_main.tex, right?

PR #100, opened 2017-08-04
Ring memory logging

jaycedowell:
Added support for logging of ring metadata (space, alignment, size, etc.) and updated several of the tools to use the new information.  This includes:
  * a new `ring` directory under `/dev/shm/bifrost/<pid>` where the metadata is stored
  * per-ring ProcLog metadata indexed by the ring name that includes:
    * space - where the ring is allocated
    * binding - NUMA binding, if applicable
    * alignment - allocation alignment, in bytes
    * ghost - size of the ghost region, in bytes
    * span - size of a span, in bytes
    * stride - stride size, in bytes
    * nringlet - the number of ringlets in use
  * updates to `like_ps.py` and `like_top.py` to work with the new structure in `/dev/shm`
  * a new script called `like_pmap.py` that provides some insight into memory usage in a pipeline
  * update to `bifrost.ring.SpanBase.data_view()` to support non-NumPy data types, i.e., 'ci4'
  * small changes to some of the "like" script to help suppress VMA messages
  * added a `core` keyword to the `bifrost.ring[2].Ring.__init__()` to set the NUMA binding for a ring


jaycedowell:
I'm not sure where the travis errors are coming from.  Do they look familiar to anyone?

benbarsdell:
The ctypesgen step failed (visible in the hidden build log), but it doesn't stop the build (I'll try to fix this soon).

Can't include string in memory.h (it has to be pure C).

src/memory.cpp
```diff
@@ -75,6 +75,22 @@ BFstatus bfGetSpace(const void* ptr, BFspace* space) {
...
 }
 
+const char* bfGetSpaceString(BFspace space) {
+	// TODO: Is there a better way to do this that does not involve hard 
+	//       coding all of these values twice (one in memory.h for the 
+	//       enum, once here)?
+	
+	
+	switch( space ) {
+		case 0:	 return "auto";
```
benbarsdell: These can use the BF_SPACE_* enumerations instead of integers.

src/ring_impl.cpp
```diff
@@ -199,6 +218,23 @@ void BFring_impl::resize(BFsize contiguous_span,
 	_span       = new_span;
 	_stride     = new_stride;
 	_nringlet   = new_nringlet;
+	
+#if BF_NUMA_ENABLED
+	_size_log.update() << "space     : " << bfGetSpaceString(_space) << "\n"
```
benbarsdell: Can this and the similar code at line 109 be moved into a function to avoid duplication? The BF_NUMA_ENABLED condition could also be made tighter to further reduce duplication.

src/ring_impl.cpp
```diff
@@ -104,6 +104,25 @@ BFring_impl::BFring_impl(const char* name, BFspace space)
 	BF_ASSERT_EXCEPTION(space==BF_SPACE_SYSTEM,
 	                    BF_STATUS_INVALID_ARGUMENT);
 #endif
+
+#if BF_NUMA_ENABLED
+	_size_log.update("space     : %s\n"
```
benbarsdell: See comment at line 223.

tools/like_pmap.py
```diff
@@ -0,0 +1,306 @@
...
+import sys
+import getopt
+import subprocess
+
+os.environ['VMA_TRACELEVEL'] = '0'
+from bifrost.proclog import load_by_pid
+
+
+def usage(exitCode=None):
+	print """%s - Get a detailed look at memory usage in a bifrost pipeline
```
benbarsdell: Need to change tabs to spaces in this file.

PR #103, opened 2017-08-08
GPU Unpack, take three

jaycedowell:
Add support for GPU-based unpacking and quantization (third attempt).  This includes:
  * GPU functions for unpacking integer data into 'i8'/'ci8'
  * Functions to unpack integer data and promote to float/double
  * Support for unpacking 1-bit data as:
     * 0 unpacks to -1
     * 1 unpacks to 1
* Preliminary support for packing into 'i2'/'ci2' and 'i1'/'ci1'.  Full support held up by a check for padded arrays.


jaycedowell:
This approach to building the unpacking and quantization modules works on ADP but it seems to cause problems with travis.  Is there a more robust way to deal with this that doesn't lead to maintaining two sets of files?

benbarsdell:
Would passing "-x c++" to g++ help?

jaycedowell:
I had something similar to the "-x c++" before that did work right either.  Maybe I should revisit this approach before splitting things up.

src/unpack.cu
```diff
@@ -0,0 +1,689 @@
...
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <bifrost/unpack.h>
```
MilesCranmer: Since so much of this file is similar, do you think you could redo this with a "git mv" (followed by the patches) so that the version history is preserved? Otherwise, git thinks the file was written from scratch, and git log/bisect/blame/etc., all lose their effectiveness.

jaycedowell: I don't think I have enough "git foo" to do that without making a huge mess.  However, I'm not ruling out the possibility of a "take four".

jaycedowell:
What's up the these travis builds?  The "details" link indicates that they have finished but here they are listed as still pending.

src/unpack.cu
jaycedowell: I'm not sure that it possible to maintain the .cpp extension given how the file is currently built.  It uses the .cu so that nvcc builds it for GPU-enabled installations.  Maybe @benbarsdell 's second suggest works with this?

Also, doesn't `git log --follow` work to preserve history across renames?

MilesCranmer:
I clicked "restart build" just now and it passed. Not sure what that was - it could be because I was editing the travis settings at the same time as the build.

PR #104, opened 2017-08-08
FIR filter

jaycedowell:
Add support for finite impulse response (FIR) filters on the GPU.


jaycedowell:
The sky is my unit test.

python/bifrost/fir.py
```diff
@@ -0,0 +1,60 @@
...
+
+from libbifrost import _bf, _check, _get, _fast_call, _string2space
+from ndarray import asarray, zeros
+
+import ctypes
+import numpy as np
+
+class FIR(object):
+    def __init__(self):
+        self.obj = _get(_bf.FIRCreate(), retarg=0)
```
benbarsdell: This can use the new BifrostObject like here:
https://github.com/ledatelescope/bifrost/blob/master/python/bifrost/fft.py#L32-L34

src/fir.cu
```diff
@@ -0,0 +1,406 @@
...
+		BF_TRACE_STREAM(_stream);
+		BF_ASSERT_EXCEPTION(_coeffs != NULL, BF_STATUS_INVALID_STATE);
+		BF_ASSERT_EXCEPTION(_state0 != NULL,  BF_STATUS_INVALID_STATE);
+		BF_ASSERT_EXCEPTION(_state1 != NULL,  BF_STATUS_INVALID_STATE);
+		BF_ASSERT_EXCEPTION(out->dtype == BF_DTYPE_CF32 || \
+		                    out->dtype == BF_DTYPE_CF64,     BF_STATUS_UNSUPPORTED_DTYPE);
+		
+		BF_CHECK_CUDA_EXCEPTION(cudaGetLastError(), BF_STATUS_INTERNAL_ERROR);
+		
+#define LAUNCH_BEAMFORMER_KERNEL(IterType,OterType) \
```
benbarsdell: LAUNCH_FIR_KERNEL?

src/bifrost/fir.h
```diff
@@ -0,0 +1,66 @@
...
+                   BFsize  ncoeff,
+                   BFsize  decim,
+                   BFsize  ntime,
+                   BFsize  nstand,
+                   BFspace space,
+                   void*   plan_storage,
+                   BFsize* plan_storage_size);
+BFstatus bfFIRSetStream(BFfir       plan,
+                        void const* stream);
+BFstatus bfFIRSetCoeffs(BFfir          plan,
```
benbarsdell: If the coefficients get changed infrequently (ever?), could they instead be passed to bfFIRInit?
They could also be passed as an array on the CPU, which could then be internally copied to the GPU.

src/bifrost/fir.h
```diff
@@ -0,0 +1,66 @@
...
+#include <bifrost/array.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct BFfir_impl* BFfir;
+
+BFstatus bfFIRCreate(BFfir* plan);
+BFstatus bfFIRInit(BFfir   plan,
```
benbarsdell: It looks like ntime could be moved to bfFIRExecute instead, which would allow it to change on each call (e.g., for a smaller ntime at the end of a data stream).
ncoeff could also be absorbed into 'coeffs' if they were passed here instead of via the separate function below.
nstand could also be absorbed into 'coeffs' if we wanted to support separate coefficients for each stand.

src/fir.cu
jaycedowell: Ha ha, yes, the FIR module is based on my beam former module.

src/bifrost/fir.h
jaycedowell: I'm not sure how frequently things would change.  The major use cases I see for this are (1) bandpass shaping and (2) decimation.  For (1) there would really only ever be one set of coefficients.  For (2) I could imagine on-demand changes to the decimation for processing, each requiring it's own set of filter coefficients.

jaycedowell: I guess there is also a third case where you have a time domain instrument and you use FIR filters to create fractional sample delays.  There you would need coefficient sets of each fractional delay and those would be externally triggered to update.  This is similar to the scheme employed in the LWA1 DP system for the beam former delays.

src/fir.cu
```diff
@@ -0,0 +1,406 @@
...
+__global__ void fir_kernel(int                   ncoeff,
+                           int                   decim, 
+                           int                   ntime, 
+                           int                   nstand,
+                           double*               coeffs,
+                           Complex64*            state0,
+                           Complex64*            state1,
+                           InType*               d_in,
+                           OutType* __restrict__ d_out) {
+	int t0 = threadIdx.x + (blockIdx.x + blockIdx.y*gridDim.x)*blockDim.x;
```
benbarsdell: Parallelising this kernel over stands as well, using the thread.x axis, would allow all of the reads and writes to be coalesced, which should give a significant speed improvement. The reads/writes would have to operate on a whole JonesVec data type (instead of looping over pol) to get the most out of it.

src/bifrost/fir.h
jaycedowell: I could see wanting to be able to set different coefficients for different stands.  That is what would be needed for the above time domain beam former concept.  I'm not sure I follow to change to ncoeff.  I think you would need to know that at initialization time so that you can setup the proper input and output buffering needed for the filter.  Am I missing something here?

src/fir.cu
jaycedowell: I initially went with this approach on ADP because the first usage case had ntime >> nstand * npol.  Is there a good thread/block division rubric that can be used to automatically tune the setup?  Also, is there another way to deal with one dimension having a large number of elements?

src/fir.cu
```diff
@@ -0,0 +1,406 @@
...
+	}
+};
+
+BFstatus bfFIRCreate(BFfir* plan_ptr) {
+	BF_TRACE();
+	BF_ASSERT(plan_ptr, BF_STATUS_INVALID_POINTER);
+	BF_TRY_RETURN_ELSE(*plan_ptr = new BFfir_impl(),
+	                   *plan_ptr = 0);
+}
+// **TODO: Passing 'BFarray const* in' here could replace nchan, f0, df and space if BFarray included dimension scales
```
benbarsdell: These remnant comments can be deleted.

src/fir.cu
```diff
@@ -0,0 +1,406 @@
...
+	BF_TRY_RETURN(plan->reset_state());
+}
+BFstatus bfFIRExecute(BFfir          plan,
+                      BFarray const* in,
+                      BFarray const* out) {
+	BF_TRACE();
+	BF_ASSERT(plan, BF_STATUS_INVALID_HANDLE);
+	BF_ASSERT(in,   BF_STATUS_INVALID_POINTER);
+	BF_ASSERT(out,  BF_STATUS_INVALID_POINTER);
+	BF_ASSERT( in->shape[ in->ndim-3] == plan->ntime(),      BF_STATUS_INVALID_SHAPE);
```
benbarsdell: Need assertions on ndim before checking shapes.

jaycedowell:
Are there any lingering issues with this?

PR #107, opened 2017-10-10
GPU Unpack, take four

jaycedowell:
Add support for GPU-based unpacking and quantization (fourth attempt). This includes:
* GPU functions for unpacking integer data into 'i8'/'ci8'
* Functions to unpack integer data and promote to float/double
* Support for unpacking 1-bit data as:
  * 0 unpacks to -1
  * 1 unpacks to 1
* Preliminary support for packing into 'i2'/'ci2' and 'i1'/'ci1'. Full support held up by a check for padded arrays.

### DIFF
--- a/python/bifrost/DataType.py
+++ b/python/bifrost/DataType.py
@@ -132,9 +132,10 @@ class DataType(object):
             self._kind, self._nbit, self._veclen = t
         else:
             t = np.dtype(t) # Raises TypeError if t is invalid
-            if t.ndim == 0:
+            ndim = getattr(t, 'ndim', len(t.shape))
+            if ndim == 0:
                 self._veclen = 1
-            elif t.ndim == 1:
+            elif ndim == 1:
                 self._veclen = t.shape[0]
                 t = t.base
             else:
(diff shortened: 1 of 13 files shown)


jaycedowell:
This is like take three but puts the new GPU quantizer/unpacker into .cu and .hu files to preserve the history in `quantize.cpp` and `unpack.cpp`.  This also adds additional unit tests for the GPU functions and the promotion functions.

MilesCranmer:
Looks good to me! @benbarsdell do you have any issues with this?

benbarsdell:
I'm not sure what happened to the CI tests here; any ideas?. All tests pass when I run this branch locally.

jaycedowell:
I'm not sure what is going on here.  I just triggered a build of master and ended up with the same error that kills this PR.  I think the culprit may be:

`test_block.py:315: UserWarning: loadtxt: Empty input file: ".log.txt"
  result = np.loadtxt(self.logfile).astype(np.float32).view(np.complex64)`

although I don't fully understand what this means.

jaycedowell:
That's interesting.  I restarted a build from the jaycedowell-gpu-unpack2 branch that was previously marked as successful and it failed.

jaycedowell:
Ah, it's the NumPy version.  I rolled back the NumPy version in the `.travis.yaml` file from the latest (1.14.3) to 1.13.3 using the jaycedowell-travistest branch and the build works.

jaycedowell:
Making sure that files are closed after `np.savetxt()` is called with a file handle fixes the testing problem.

benbarsdell:
Thanks for tracking that down Jayce!

PR #126, opened 2019-06-13
Multi-logging fixes

jaycedowell:
A couple of fixes for the ProcLog module.  The first is to wrap the `/dev/shm/bifrost/` directory cleanup in a try...catch block to deal with files owned by other users and their pipelines.  The second is a change to the ProcLog name de-duplicator to make it work with the `module/file` structure that is currently used so that duplicate modules use separate directories.


jaycedowell:
No comments in over three weeks, merging.

PR #127, opened 2019-06-13
Quantize/unpack stream fix

jaycedowell:
Updated the GPU versions of quantize and unpack so that they run in `g_cuda_stream` instead of the default stream.  Also, enabled BF_TRACE and BF_TRACE_STREAM in these two modules where relevant.


jaycedowell:
No comments in over three weeks, merging.

PR #132, opened 2019-10-02
Map disk cache

jaycedowell:
Adds a disk-based caching system for `bifrost.map()` calls to help reduce start up time in pipelines that are run in the same configuration multiple times.  This feature can also be turned off by setting the "NOMAPCACHE" variable to one in the `user.mk` file.


jaycedowell:
Oh, the "coverage" thing again.

telegraphic:
+1 for this functionality -- as this is C++ level I would defer to others for review though

jaycedowell:
The cache should be controllable through `configure` now.

jaycedowell:
I think this PR is probably ready for someone to look at it again, pending tests of whether or not this still works on a GPU.

src/fileutils.hpp
```diff
@@ -0,0 +1,141 @@
...
+	if( std::system(("rm -f "+path).c_str()) ) {
+		throw std::runtime_error("Failed to remove file: "+path);
+	}
+}
+inline bool file_exists(std::string path) {
+    struct stat s;
+    return !(stat(path.c_str(), &s) == -1
+	         && errno == ENOENT);
+}
+bool process_exists(pid_t pid) {
```
league, line 74: Although this function is long, it either needs to be marked `inline` like the others or moved out into a single `.cpp`. I'm getting a link error because it's duplicated in `map.o` and `proclog.o`:

```
map.o: In function `process_exists(int)':
/home/qblocks/bifrost/src/fileutils.hpp:74: multiple definition of `process_exists(int)'
proclog.o:/home/qblocks/bifrost/src/fileutils.hpp:74: first defined here
```

Will try a fix locally and see what happens next…

test/test_map.py
```diff
@@ -36,6 +36,7 @@
 class TestMap(unittest.TestCase):
     def setUp(self):
         np.random.seed(1234)
+        bf.map_clear_cache()
```
league, line 39: Everything in `test_map` and `test_transpose` bombed because `map_clear_cache` is not exported properly:
```
ERROR: test_complex_float (test_map.TestMap)
----------------------------------------------------------------------
  File "/home/qblocks/bifrost/test/test_map.py", line 39, in setUp
    bf.map_clear_cache()
AttributeError: module 'bifrost' has no attribute 'map_clear_cache'
```
I can see `def clear_cache` in `bifrost/map.py` but it's not exported by `bifrost/__init__.py`

python/bifrost/map.py
```diff
@@ -142,3 +145,31 @@ def map(func_string, data, axis_names=None, shape=None,
...
+        except OSError:
+            pass
+            
+    if len(output) == 0:
+        output = 'none'
+    print(output)
+
+
+def clear_cache():
+    _check(_bf.bfMapClearCache())
```
league, line 175: I also tried building this branch with `--disable-map-cache`, and adjusted the imports and naming so that `map_clear_cache` in the test `setUp` comes here. But then this `_check` fails with `BF_STATUS_UNSUPPORTED` (because map cache is turned off). So we need to decide if it's okay to call clear_cache if cache is disabled... one POV is that such an operation could alway succeed? ("Yep, cache is clear. Because there isn't one!") The alternative, I guess, is to guard against calling it here, or in the `setUp` methods. But fewer conditionals are better than more. :)

league (commented):
After minimal fixes to the issues I noted inline, the `test_transpose` and `test_map` succeed. These may not be the most comprehensive or sensible fixes for these, but for the record, they were:

```diff
--- a/python/bifrost/__init__.py
+++ b/python/bifrost/__init__.py
@@ -40,3 +40,3 @@ from bifrost.ndarray import ndarray, asarray, empty_like, empty, zeros_like, zer
 from bifrost import views
-from bifrost.map import map
+from bifrost.map import map, clear_cache
 from bifrost.pipeline import Pipeline, get_default_pipeline, block_scope
diff --git a/src/fileutils.hpp b/src/fileutils.hpp
index 637b6b3..a4790fe 100644
--- a/src/fileutils.hpp
+++ b/src/fileutils.hpp
@@ -73,3 +73,3 @@ inline bool file_exists(std::string path) {
 }
-bool process_exists(pid_t pid) {
+inline bool process_exists(pid_t pid) {
 #if defined __APPLE__ && __APPLE__
diff --git a/test/test_map.py b/test/test_map.py
index 2dfcdbb..8c24980 100644
--- a/test/test_map.py
+++ b/test/test_map.py
@@ -38,3 +38,3 @@ class TestMap(unittest.TestCase):
         np.random.seed(1234)
-        bf.map_clear_cache()
+        bf.clear_cache()
     def run_simple_test(self, x, funcstr, func):
diff --git a/test/test_transpose.py b/test/test_transpose.py
index b1083f3..6780c59 100644
--- a/test/test_transpose.py
+++ b/test/test_transpose.py
@@ -38,3 +38,3 @@ class TransposeTest(unittest.TestCase):
     def setUp(self):
-        bf.map_clear_cache()
+        bf.clear_cache()
     def run_simple_test(self, axes, dtype, shape):
```

There's still some debugging/tracing output:
```
CACHE: /home/qblocks/.bifrost/map_cache/
HOME: /home/qblocks
```

src/fileutils.hpp
league: I see now that most of the functionality now in `fileutils.hpp` was originally in `proclog.o` and not inline. If you like, I can try to figure out how to keep it in a (possibly new) `.cpp` but refer to it from both places. (Rather than inline.)

jaycedowell: Yes, it was in `proclog.cpp` before but I wanted to use it in `map.cpp` as well.  `process_exists` also used to be a lot simpler when it was linux-only.  The mac support is what caused it to balloon.

test/test_map.py
jaycedowell: Ah, yes, a last minute change in the naming inside `bifrost.map`.  I didn't really like `map_clear_cache` so I went with a simpler `clear_cache`.  Maybe a `clear_map_cache` and `list_map_cache` would be better?  That seems a little more readable once they are imported into `bifrost/__init__.py` than a simple `clear_cache`.  Clear what cache?

python/bifrost/map.py
jaycedowell: Yeah, I think always returning `BF_STATUS_SUCCESS` would be better approach if the cache is disabled.  That style is what I used for the cache listing function in `bifrost.map`.

test/test_map.py
league: Yep. As long as we're keeping flat-ish namespace, `bifrost.clear_map_cache` (and `list_map_cache`) has my vote. If we wanted to get hierarchical, then `bifrost.map.clear_cache` but that doesn't seem worthwhile.

league:
One other thing I thought to look into… calling `clear_cache` in the test `setUp` is appropriate to reduce test dependence and ordering effects. But we should make sure testing includes the possibility of a cache hit, in case that path is (or becomes) buggy.

jaycedowell:
My original motivation for using so much `clear_cache` in the tests was to make sure that we didn't end up with failures due to cache hits on things used in a previous version of Bifrost.  Maybe we should only be calling it once at the start of `test_map`/`test_transpose`.

test/test_transpose.py
```diff
@@ -35,8 +35,12 @@
 
 @unittest.skipUnless(BF_CUDA_ENABLED, "requires GPU support")
 class TransposeTest(unittest.TestCase):
+    first_test = True
+    
     def setUp(self):
-        bf.map_clear_cache()
+        if self.first_test:
```
league, line 41: There's a `@classmethod setUpClass` that would probably be good for this, but it originates with python 3.2. https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

src/map.cpp
```diff
@@ -796,6 +793,6 @@ BFstatus bfMapClearCache() {
     DiskCacheMgr::get().clear();
     return BF_STATUS_SUCCESS;
```
league, line 794: Suggest move the return outside the if, then don't need else?

test/test_transpose.py
jaycedowell: All the great things that we will have when we stop supporting Python2.7...

league: A frabjous day that shall be. Could start hash-tagging places to revisit once free, haha.

jaycedowell:
I did some testing today and fixed a few problems that I found.  Everything is now working as far as I can.  Nothing happens with regards to the cache if you compile with `--disable-map-cache`.  The cache gets cleared when the tests start.  The cache doesn't get cleared if you comment out the call to `clear_map_cache`.  When the cache clearing isn't called the tests run faster the second time.  The only outstanding issue that I know of is the `fileutils.hpp`/`process_exists()` one.  Additional testing with another version of CUDA (I used 10.1) would also be nice.

jaycedowell:
I did some testing today and fixed a few problems that I found.  Everything is now working as far as I can.  Nothing happens with regards to the cache if you compile with `--disable-map-cache`.  The cache gets cleared when the tests start.  The cache doesn't get cleared if you comment out the call to `clear_map_cache`.  When the cache clearing isn't called the tests run faster the second time.  The only outstanding issue that I know of is the `fileutils.hpp`/`process_exists()` one.  Additional testing with another version of CUDA (I used 10.1) would also be nice.

league:
I should be able to try this PR again today, on cuda 11. I'll also start to look at one of the others, whatever is most relevant or ready: is that mmap ring space #135 ?

jaycedowell:
Cool, thanks.  I just finished going through all of my tests (again) under 10.1 and things still seem ok.

If you did want to look at another one either #135 or #167 would work.  #167 is short so it has that going for it.

league:
Tests worked great for me on cuda11, tried with and without map cache enabled. It makes a huge difference in speed. Just out of curiosity, I messed with overwriting one of the files with random bytes and sure, it core dumped. Not surprising, and not worth detecting… just trying to push the limits, find the "robustness boundary." Then I tried editing one of the numbers in `cache.version`, and it seemed to **hang** instead of dump. (But that's with the corrupt file still there. Not sure whether that's significant. If the version doesn't match expectations, we shouldn't be trying to read one of the other files?)

If I delete the whole directory and start over, it works again of course. Then if I change `cache.version` _without_ also corrupting one of the other files, it seems to replace it cleanly and keep moving. So I think that's okay.

I started to experiment with adding a `fileutils.cpp` in addition to the `hpp` but it's not ready yet, and maybe not necessary. I'm also slightly uncomfortable with dumping together a command line and calling `system()` in these routines, but I get it… it's definitely easier to use `mkdir -p` and `rm -rf` and `*.inf` than to code those things in C++ (or maybe even to find the libraries to do it).

league:
I remember seeing an issue about machines with >1 GPU, and I'm wondering if the map cache makes that any worse… is it embedding GPU-0 architecture info. Or what if $HOME is shared across machines with different GPUs. Maybe not worth worrying about, since if something goes wrong it's a simple matter to wipe the directory and start fresh. Just thinking out loud about potential complications.

jaycedowell:
> Then I tried editing one of the numbers in `cache.version`, and it seemed to **hang** instead of dump. (But that's with the corrupt file still there. Not sure whether that's significant. If the version doesn't match expectations, we shouldn't be trying to read one of the other files?)

The hang is surprising.  The `cache.version` should have been verified first and then the entire cache should have be dropped when the version information was wrong.

> I started to experiment with adding a `fileutils.cpp` in addition to the `hpp` but it's not ready yet, and maybe not necessary. I'm also slightly uncomfortable with dumping together a command line and calling `system()` in these routines, but I get it… it's definitely easier to use `mkdir -p` and `rm -rf` and `*.inf` than to code those things in C++ (or maybe even to find the libraries to do it).

I had thought about this too.  The options are:
 * move to C++17 and use the new filesystem library,
 * stay with C++11 and use boost for their `boost::filesystem` library (it's what the C++17 one is based on),
 * use C system calls to implement `mkdir`/`rm` and globbing, or
 * stick with `system()`.

jaycedowell:
> I remember seeing an issue about machines with >1 GPU, and I'm wondering if the map cache makes that any worse… is it embedding GPU-0 architecture info. Or what if $HOME is shared across machines with different GPUs.

It shouldn't make things worse on multi-GPU systems but it might be problematic if those GPUs are heterogeneous.  The cache is stored as PTX which should be forward compatible.  The hitch is that when `bifrost.map` generates the PTX it targets the architecture of the GPU it has been requested on and then that will get saved to the disk cache.  On a mixed GPU system you could end up in a situation where you cache a function for a compute capability 75 device, save it as 75, and then try to load it/use it from a 52 device.  I don't think that works.

The alternative would be to do something like always generate the PTX for the lowest compute capability that Bifrost was compiled for if the disk cache is enabled.  That would solve my above problems, maybe with a little slowdown, but be largely unchanged in the homogeneous GPU case.

league:
- Regarding `cache.version`, yeah I'm not sure. It was a hard enough hang that `^C` didn't do the trick, had to kill by PID. Could explore/trace more carefully if it's worrisome.
- Regarding C++ filesystem: there seems to still be a lot of Ubuntu 18 out there. If there's not a C++17-supporting gcc in its regular apt sources, it might be too early to make that switch. Boost may be plausible, and especially if it wouldn't take major changes to reroute it to C++ standard stuff later on. “Hard pass” on implementing `mkdir -p`, `rm -rf` and glob on our own in C++.
- Regarding GPU heterogeneity, it's likely an obscure situation. But would it be relatively easy to – rather than encode the architecture in `cache.version`, could you use a separate subdir per architecture/CUDA version/whatever, like `$HOME/.bifrost/map_cache/ARCH-VERSION-TAG/*.ptx`. Then if you end up on a different architecture, the appropriate cache would just appear to be absent, rather than existing but with the wrong arch. Does that make sense?

I don't think any of this necessarily has to hold up this branch. Could leave a TODO here or there as a reminder for things to pursue in the future or revisit if there's a relevant bug report. Or write it as a low-priority github issue with appropriate label (wish list, improvement, for future consideration, etc.)?

jaycedowell:
Your point on splitting up the cache by arch. does make sense but I'm not sure how necessary that is if we stick with PTX for the cache.  As far as I can tell PTX gets interpreted by the CUDA driver and JIT compiled to a binary when it is run.  So as long as we aren't using features in the PTX that the card cannot support it should be ok to have a flat cache.  That should be true if we always target the lowest arch. requested during the install.  Getting fancier is an option but I think how we do that is going to need some more thought about the details of compatibility and what CUDA really does with PTX files/binaries.  Maybe a "TODO" is the way to go.

league:
Diagnosed the hang that I experienced before. After it core-dumped because I corrupted a ptx file, it left behind the `.lock` file. So next time when trying to open the cache, it was waiting on the lock. For whatever that's worth. I was messing around and probably got what I deserved. :)

jaycedowell:
I ran through my tests again this morning and I think I'm happy with this.  @league if you are happy with it too then I will merge it.

league:
LGTM, ship it. 🚢

PR #133, opened 2019-10-15
Testbench and Tools Modernization

jaycedowell:
Work on modernizing the testbench and tools scripts by converting over to print_function and argparse.  Also, added a simple test suite that uses pylint to give the scripts a basic sanity check during testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: python
 python:
     - 2.7
     # PyPy versions
-    - pypy-5.4.1
+    - pypy2.7-6.0
 
 services:
     - docker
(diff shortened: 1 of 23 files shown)


telegraphic:
Had a quick (i.e. not comprehensive) look through and think this all looks great. OK to merge IMO

jaycedowell:
test_scripts.py runs a static analysis of the scripts using pylint to try and find obvious errors.  It is not as good as actually running the script with data and comparing the output but I think that it is better than nothing.

telegraphic:
Ah ok, well still LGTM.

jaycedowell:
Let's do this!

PR #135, opened 2020-02-03
mmap()'d Ring Space

jaycedowell:
Adds a new mmap()'d space for Bifrost rings that allows for larger disk-based rings for buffering.


jaycedowell:
I'm generally happy with how it works but the specification of the base directory in `user.mk` seems a little fiddly.  I'm also not sure how to allow multiple base directories to be specified.

jaycedowell:
I think the environment variable approach is viable and  takes care of my last concern.

jaycedowell:
Well, at least mapped compiles on MacOS now.  That's something.

jaycedowell:
Keep the logging directory and the mapped ring directory separate on mac seems to have helped with a lot of the testing failures but not there is a segfault when the test suite exits.

jaycedowell:
Looks like there are some problems related to the recent changes to `filesystem.[ch]pp`.

Issue #115, opened 2018-03-27
ValueError: invalid literal for long() with base 8: '08420217248550443400745280086994171'


gijzelaerr:
compiling latest master branch 1e61eda4276f5cebf3784234df333f39ba947612 gives
```
λ  make
make -C src all
make[1]: Entering directory '/packaging/kern/upstream_repos/bifrost/src'
make[1]: Leaving directory '/packaging/kern/upstream_repos/bifrost/src'
make -C python build
make[1]: Entering directory '/packaging/kern/upstream_repos/bifrost/python'
ctypesgen.py -lbifrost -I../src ../src/bifrost/address.h ../src/bifrost/ring.h ../src/bifrost/cuda.h ../src/bifrost/fdmt.h ../src/bifrost/transpose.h ../src/bifrost/udp_capture.h ../src/bifrost/memory.h ../src/bifrost/udp_transmit.h ../src/bifrost/common.h ../src/bifrost/unpack.h ../src/bifrost/udp_socket.h ../src/bifrost/affinity.h ../src/bifrost/fft.h ../src/bifrost/map.h ../src/bifrost/proclog.h ../src/bifrost/linalg.h ../src/bifrost/array.h ../src/bifrost/quantize.h ../src/bifrost/reduce.h -o bifrost/libbifrost_generated.py
INFO: Status: Preprocessing /tmp/tmpzPyj7b.h
INFO: Status: gcc -E -U __GNUC__ -dD -I../src "-Dinline=" "-D__inline__=" "-D__extension__=" "-D__const=const" "-D__asm__(x)=" "-D__asm(x)=" "-DCTYPESGEN=1" "/tmp/tmpzPyj7b.h"
Traceback (most recent call last):
  File "/usr/bin/ctypesgen.py", line 154, in <module>
    descriptions=ctypesgencore.parser.parse(options.headers,options)
  File "/usr/lib/python2.7/dist-packages/ctypesgencore/parser/__init__.py", line 21, in parse
    parser.parse()
  File "/usr/lib/python2.7/dist-packages/ctypesgencore/parser/datacollectingparser.py", line 69, in parse
    ctypesparser.CtypesParser.parse(self, fname, None)
  File "/usr/lib/python2.7/dist-packages/ctypesgencore/parser/cparser.py", line 113, in parse
    self.preprocessor_parser.parse(filename)
  File "/usr/lib/python2.7/dist-packages/ctypesgencore/parser/preprocessor.py", line 200, in parse
[...]

PR #136, opened 2020-02-03
Python 3 Support and Docker Update

jaycedowell:
This PR adds support for Python 3 through a combination of changes to Bifrost itself as well as a new version of ctypesgen (which also addresses #115).  It also updates the Docker GPU image to target Ubuntu 18/CUDA 10.2 (to address #129) and adds the tools in `tools` to the Python installation.


python/bifrost/addon/leda/blocks.py
```diff
@@ -28,6 +28,11 @@
 This file contains blocks specific to LEDA-OVRO.
 """
 
+from __future__ import print_function
+import sys
+if sys.version_info > (3,):
```
telegraphic, line 33: Might be slightly cleaner to use `six` -- https://six.readthedocs.io/#module-six.moves

jaycedowell: Yeah, I was trying to avoid adding a dependency.  At some point this should all go away as well since Python2 will be forgotten.

telegraphic: Fair enough! This time last year I thought Py2.7 would end up like FITS and still be used in 50 years, but looks more like Py3 is finally prevailing

jaycedowell: Ha ha, yeah, I was happy to stick with Py2 until major packages started dropping support.

Looking forward I should probably switch the logic of the "if" statements, i.e., `if sys.version_info <= (3,):` to make it easier to strip out later.

jaycedowell:
Thanks for taking a look at everything.  The "fir" and "memory" changes were in response to some compiler warnings I started seeing under newer versions of Ubuntu/CUDA.  It's nothing shocking but I should probably run the test suite on "fir" locally just to make sure it all checks out.

jaycedowell:
Ok, I think this is much better now.  I've also setup a Jenkins server here at UNM which enables me to run the almost all of the test suite that uses the GPU on both Python 2.7 and Python 3.6.  The only thing that I haven't been able to run is the FDMT test suite.  I also might look into @telegraphic  's idea of running the scripts under `testbench` for added confidence.

benbarsdell:
What's the issue with the FDMT test suite?

jaycedowell:
When I was initially getting things running on Jenkins I was running out of memory on the FDMT test.  Maybe it is worth trying again now that everything else is working.

jaycedowell:
Huh, the FDMT test seems to work now (and passes).  Yay.

jaycedowell:
@telegraphic I'm having problems running some of the "testbench" examples.  `test_guppi.py` and `test_guppi_reader.py` don't seem to do anything.  `test_fft_detect.py` initially would not run and still throws errors after I tweaked it.

jaycedowell:
Does anyone have any final comments on this PR or opinions about updating the Bifrost version to 0.9.0?  If so, comment now or forever hold your peace.

PR #139, opened 2020-04-15
Merging upstream + dp4a codes

telegraphic:


telegraphic:
please ignore -- should be on my local branch only (not ready to push upstream!)

FWIW -- this is a small dp4a correlator (up to N=32), and beanfarmer CUDA beamformer port

PR #153, opened 2021-10-20
Merge of telegraphic/main and ledatelescope/autoconf

telegraphic:
Hi Jayce,

I just merged the the autoconf branch with my repo, which was moderately successful (most unit tests still pass, and the ones that don't didn't before! I'm particularly worried by the non-passing FFT tests).

One thing that needs sorting is that I was using `user.mk` to define whether or not to build certain codes. That is, in `src/Makefile` I have:

```
ifndef NOUDPSOCKET
  LIBBIFROST_OBJS += \
    address.o \
    udp_socket.o \
  udp_capture.o \
  udp_transmit.o
endif

ifdef BUILDDP4A
  LIBBIFROST_OBJS += \
  xcorr_lite.o\
  beanfarmer.o
endif
```

So that these codes aren't built unless the user enables them (at Molonglo the Linux kernel is old on some machines, so didn't support some of the `recvmsg` UDP socket code!).

I couldn't get `./configure; make` to run unless I manually exported `$GPU_ARCH`, which was previously set in `user.mk`. What's the best way around this?

By the by, I feel that the DP4A-specific codes would be better as plugins, and out of the core bifrost repo. Probably the romein kernels too, and perhaps the FDMT?


jaycedowell:
I'm closing this for now but @league and I need to take a closer look at these changes to see what is still valid now that v0.10.0 has been released.

Issue #154, opened 2021-10-26
Problems encountered while installing and running 'make -j'


yushilingco:
$ make -j
udp_capture.cpp:87:32: fatal error: mellanox/vma_extra.h: No such file or directory
compilation terminated.
autodep.mk:32: recipe for target 'udp_capture.o' failed
make[1]: *** [udp_capture.o] Error 1
...
ptxas /tmp/tmpxft_000029f6_00000000-7_fft.compute_35.ptx, line 2379; error   : Illegal operand type to instruction 'ld'
ptxas /tmp/tmpxft_000029f6_00000000-7_fft.compute_35.ptx, line 2379; error   : Unknown symbol '__unnamed_1_param_0'
ptxas /tmp/tmpxft_000029f6_00000000-7_fft.compute_35.ptx, line 2747; error   : Illegal operand type to instruction 'ld'
ptxas /tmp/tmpxft_000029f6_00000000-7_fft.compute_35.ptx, line 2747; error   : Unknown symbol '__unnamed_2_param_0'
ptxas fatal   : Ptx assembly aborted due to errors
autodep.mk:56: recipe for target 'fft.o' failed
make[1]: *** [fft.o] Error 255
make[1]: Leaving directory '/home/yushiling/software/bifrost/bifrost-master/src'
Makefile:15: recipe for target 'libbifrost' failed
make: *** [libbifrost] Error 2

I encountered the above problems during installation. I tried to install the missing package, but it seems incomplete. I hereby ask for the reason. Thank you very much.

yushilingco:
I'm sorry I can't find GPU_ARCHS_VALID and GPU_ARCH_LATEST. And can you provide a configure for testing ？Thanks a lot.

jaycedowell:
How GPU_ARCHS_VALID and GPU_ARCH_LATEST are defined is in `src/Makefile`.  All you should need to do is convert the make shell calls into normal shell calls.  Something like the following bash script:
```
#!/bin/bash

GPU_ARCHS=86
GPU_ARCHS_SUPPORTED=`nvcc -h | grep -Po "compute_[0-9]{2}" | cut -d_ -f2 | sort | uniq`
GPU_ARCHS_VALID=`echo "${GPU_ARCHS} ${GPU_ARCHS_SUPPORTED}" | xargs -n1 | sort | uniq -d | xargs`
[...]

Issue #147, opened 2020-10-15
Segfault in Python 3


telegraphic:
Not sure if this is two separate issues, or one inter-related.

### Part 1

I was trying to get bifrost working in Python 3.7, and ran into PEP479, which changes how `StopIteration` is handled in a generator:
https://www.python.org/dev/peps/pep-0479/

Basically `StopIteration` now gets converted into a `RuntimeError`.  My workaround was instead to raise a new exception, `EndOfDataStop`, and catch that in any code that uses `yield` or explicitly catches `StopIteration`. This seems to work, athough not sure it is the optimal approach.

### Part 2

After fixing the `StopIteration` issue, I'm now running into an issue where the code crashes upon pipeline shutdown, with an error:

```
*** glibc detected *** python: double free or corruption (fasttop): 0x00007f5600006460 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x3d1ce75dee)[0x7fd5d35a7dee]
/lib64/libc.so.6(+0x3d1ce78c3d)[0x7fd5d35aac3d]
/home/dprice/install/bifrost/lib/libbifrost.so.0(+0x5b84d)[0x7fd5ca1f684d]
/home/dprice/install/bifrost/lib/libbifrost.so.0(+0x5376c)[0x7fd5ca1ee76c]
/home/dprice/install/bifrost/lib/libbifrost.so.0(bfRingDestroy+0x13)[0x7fd5ca1dea33]
/home/dprice/mpy36/lib/python3.6/lib-dynload/../../libffi.so.7(+0x69dd)[0x7fd5d3cf59dd]
/home/dprice/mpy36/lib/python3.6/lib-dynload/../../libffi.so.7(+0x6067)[0x7fd5d3cf5067]
/home/dprice/mpy36/lib/python3.6/lib-dynload/_ctypes.cpython-36m-x86_64-linux-gnu.so(_ctypes_callproc+0x2ce)[0x7fd5d3b2eede]
/home/dprice/mpy36/lib/python3.6/lib-dynload/_ctypes.cpython-36m-x86_64-linux-gnu.so(+0x13915)[0x7fd5d3b2f915]
python(_PyObject_FastCallDict+0x8b)[0x7fd5d3e18bfb]
python(+0x19abae)[0x7fd5d3ea0bae]
python(_PyEval_EvalFrameDefault+0x30a)[0x7fd5d3ec325a]
python(+0x194c1b)[0x7fd5d3e9ac1b]
[...]

Issue #134, opened 2020-01-27
Problem with sudoer,ImportError: bifrost not found


Leeying1235:
Hi,folks:
When i use bifrost with root permission,i get problems as below:
 ```
  File "test_pcap.py", line 2, in <module>
  from code.file_write import FileWriteBlock
  File "/home/ubuntu/Desktop/python_script/final_code/code/file_write.py", line 1, in <module>
    from bifrost.pipeline import SinkBlock
  File "/usr/local/lib/python2.7/dist-packages/bifrost-0.8.0-py2.7.egg/bifrost/__init__.py", line 34, in <module>
    import core, memory, affinity, ring, block, address, udp_socket
  File "/usr/local/lib/python2.7/dist-packages/bifrost-0.8.0-py2.7.egg/bifrost/core.py", line 29, in <module>
    from libbifrost import _bf
  File "/usr/local/lib/python2.7/dist-packages/bifrost-0.8.0-py2.7.egg/bifrost/libbifrost.py", line 38, in <module>
    import libbifrost_generated as _bf
  File "/usr/local/lib/python2.7/dist-packages/bifrost-0.8.0-py2.7.egg/bifrost/libbifrost_generated.py", line 609, in <module>
    _libs["bifrost"] = load_library("bifrost")
  File "/usr/local/lib/python2.7/dist-packages/bifrost-0.8.0-py2.7.egg/bifrost/libbifrost_generated.py", line 369, in load_library
    raise ImportError("%s not found." % libname)
ImportError: bifrost not found.
```
Any hints,and how to solve it?
Thanks,
Lee

jaycedowell:
It sounds like a LD_LIBRARY_PATH problem.  Have you tried updating this environment variable to include the installation path of libbifrost.so?

jaycedowell:
Closing with the release of v0.10.0.

Issue #130, opened 2019-08-12
bf.map does not work on ci32 type


telegraphic:
The complex int32 type doesn't work with the `map` function at the moment:

```python
a = bf.ndarray(np.array([1,2,3,4,5]), space='cuda', dtype='ci32')
b = bf.ndarray(np.array([1,2,3,4,5]), space='cuda', dtype='ci32')
bf.map("a = b", data={'a': a, 'b': b})
```
This raises a runtime error:
```
RuntimeError                              Traceback (most recent call last)
<ipython-input-38-faef62e8891f> in <module>()
----> 1 bf.map("a = b", data={'a': a, 'b': b})

/home/dprice/python/lib/python2.7/site-packages/bifrost-0.8.0-py2.7.egg/bifrost/map.pyc in map(func_string, data, axis_names, shape, func_name, extra_code, block_shape, block_axes)
    124                      narg, _array(args), _array(arg_names),
    125                      func_name, func_string, extra_code,
--> 126                      _array(block_shape), _array(block_axes)))

/home/dprice/python/lib/python2.7/site-packages/bifrost-0.8.0-py2.7.egg/bifrost/libbifrost.pyc in _check(status)
    101             else:
    102                 status_str = _bf.bfGetStatusString(status)
--> 103                 raise RuntimeError(status_str)
    104     else:
    105         if status == _bf.BF_STATUS_END_OF_DATA:
```
It looks like the `dtype2ctype_string` hasn't been implemented https://github.com/ledatelescope/bifrost/blob/8e1c0e13f8ba916ee1fc271943f7a2d0a836e849/src/array_utils.hpp#L63 (it's commented out).

Int32 is probably more useful now than it was previously, as the dp4a instruction takes 8-bit inputs and accumulates into a 32-bit integer. I'll see if I can get things working, but any thoughts / comments about why this was tricky / left as a todo would be appreciated.

jaycedowell:
The only thing I can think of is that there it could be related to the underlying Complex class.  If you look in [Complex.hpp](https://github.com/ledatelescope/bifrost/blob/master/src/Complex.hpp), `Complex<int>` isn't really addressed by the current structure.  It's not a floating point type but it also isn't a storage type.

jaycedowell:
While I'm at it this also reminds me of the off-and-on problems I sometimes hit with map and mixing various complex types.  It usually involves trying to work with storage types and them not getting automatically promoted when they are involved in some math.

[...]

PR #157, opened 2021-12-01
autotools-based build system

jaycedowell:
This PR switches Bifrost to an autotools-based build system would should solve many of the installation issues that people report, e.g., #109, #134, and #154.  It also adds a `.m4` and a pkg-config file to help other projects that may depend on Bifrost find an installation.

This also address a few outstanding issues with Bifrost, including:
 * support for the `cuda_managed` memory space on Pascal and later devices (#111)
 * support for the `ci32` data type in `bf.map` (#130)
 * [PEP479](https://www.python.org/dev/peps/pep-0479/) support (#147)
 * switch to an "official" version of `ctypesgen` (#161)

Finally, this PR also moves away from TravisCI in favor of GitHub Actions.


jaycedowell:
@telegraphic and @league what do you think about this?  Do you think it is ready to be merged in or does it need more work/testing?

config/cuda.m4
```diff
@@ -0,0 +1,169 @@
...
+    
+    AC_MSG_CHECKING([for valid CUDA architectures])
+    ar_requested=$( echo "$GPU_ARCHS" | wc -w )
+    ar_supported=$( ${NVCC} -h | ${GREP} -Po "'compute_[[0-9]]{2,3}" | cut -d_ -f2 | sort | uniq )
+    ar_valid=$( echo $GPU_ARCHS $ar_supported | xargs -n1 | sort | uniq -d | xargs )
+    ar_found=$( echo $ar_valid | wc -w )
+    if test "$ar_requested" = "$ar_found"; then
+      AC_MSG_RESULT([yes])
+    else
+      AC_MSG_ERROR(only architectures $ar_valid are supported)
```
league, line 154: I think this might need to be "only architectures $ar_supported" are supported.

I have occasionally (when specifying `--with-gpu-archs` incorrectly) seen the message come out as "only architectures  are supported". It looks like `$ar_supported` is what it gets from `nvcc -h`, and `$ar_valid` is the intersection of that with `$ar_requested` ?

jaycedowell: Yes, `$ar_supported` is everything supported by CUDA and `$ar_valid` is that is supported and what is requested.  I _think_ my original intention here was for this to catch and report when a partially valid list of archs. is provided.  So requesting something like "30 60" would tell you that only 60 was supported.  I hadn't thought about the case where nothing that was requested is supported.

jaycedowell:
> I'm wondering though, is it your experience that the GPU-requiring tests are reliably passing? I have a VM with GeForce RTX 3090 and CUDA 11 that succeeds through configure (it auto-detects GPU arch 80) and builds, but it's reporting a few failures.
> 
> ```
> Ran 275 tests in 713.388s
> FAILED (failures=4, errors=24, skipped=5)
> ```

I vaguely recall running into something like this in the past with regards to the FFT tests but it wasn't consistent.  Are all of your failures related to the FFT tests?  Also, what do the errors look like?

config/cuda.m4
jaycedowell: Maybe a better message would be "only '$ar_valid' of the requested archs. are supported"?

league: Sure… when things go awry, it can be useful to have both, though it may not fit as nicely into the one-liner format configure likes:
```
Doing this... result is that
```
But usually `config.log` contains necessary detail. In the cases where I got a blank output there, I think it was because the test program failed to link... that error didn't appear on stdout, but it was in the log.

jaycedowell: [54ed45b](https://github.com/ledatelescope/bifrost/pull/157/commits/54ed45bfc809f35c74f008d468d9ef39be31fbe2) tries to do both.

jaycedowell:
> This could be a really small (but compelling) use case for a reproducible build step… once we can define clearly what autoconf needs to do the right thing (which maybe won't be easy), we can leave the job of regenerating ./configure to that tool. But in general I agree it's a mistake to leave configure out of the repo and push the problem onto the user with a bootstrap script.

I looked into how I build the configure file and I use autotools installed via homebrew on my laptop.  I get around the various m4 problems with a cache of m4 files that I found I needed.  Even with all of this I still end up with "possible undefined macro: AM_DEFAULT_VERBOSITY" error which pops up from the doxygen configuration step.

league:
I don't mean to delay the autoconf merge any further, I think it could become master ASAP. But I noticed this message a lot when running `./configure`:

```
./configure: line 7404: /usr/bin/file: No such file or directory
```

It seems like the absolute path to `file` originates with `libtool`? But the `file` command is not guaranteed to be installed, it's not part of `coreutils`. And even if it is installed, it may not be in `/usr/bin`. This error occurs on colab because their ubuntu doesn't have `file` installed. The same may be true on qblocks. And on NixOS it's not part of the base install either, but even when it is installed, it won't be in `/usr/bin`. So why doesn't libtool just get it from the `PATH`? If it's a real dependency, I guess we can specify it as such, but configure shouldn't hard-code its location.

I can't tell where this error has caused any harm. I guess it's trying to determine object format, and whether it requires special linker flags... and on these 64-bit gcc/gnu/linux boxes, we don't need them.

jaycedowell:
Hrm, I hadn't noticed that `/usr/bin/file` problem.  Maybe we could do something like this?  https://lists.gnu.org/archive/html/bug-libtool/2014-06/msg00005.html

league:
Interesting how Ludovic got some push-back on that thread! (He's a Guix developer, which is very similar to NixOS). It's funny, for those systems this is pretty easily solved (if it even matters) by a patch step that would just `sed` references to `/usr/bin/file` into the appropriate hashed version in `/nix/store/zzzzz-file/bin/file`.

What was more curious/alarming to me is that I see the same message on Ubuntu. It seems that `file` is not part of a basic install, and not even part of `build-essential`. So we don't have it on Google colab or a default ubuntu docker image either:

```
$ docker run -it --rm ubuntu:latest file --version
docker: Error response from daemon: OCI runtime create failed:
container_linux.go:380: starting container process caused: exec: "file":
executable file not found in $PATH: unknown.
```

But none of these platforms seem to need those linker flags libtool is detecting, so perhaps moot point. Maybe this is just something to keep an eye on. Or just recommend `file` as one of our prerequisites: `sudo apt install exuberant-ctags file`

jaycedowell:
In the absence of a clear direction to go in maybe we just leave the `file` thing alone.

test/test_fft.py
```diff
@@ -141,7 +141,9 @@ def run_test_c2c(self, shape, axes):
         self.run_test_c2c_impl(shape, axes, inverse=True, fftshift=True)
     def run_test_c2r(self, shape, axes):
         self.run_test_c2r_impl(shape, axes)
-        self.run_test_c2r_impl(shape, axes, fftshift=True)
+        if BF_CUDA_VERSION < 10.1 and BF_CUDA_VERSION > 11:
+            # TODO: How do we announce that this is happening?
```
league, line 145: I may have an idea about that, but I need to look up some specifics about unittest and I won't get to it today. In places where we're using this pattern of `run_test_blah` which is fed different arguments by different `test_xyz` methods, there's some way to use decorators to specify what the arguments will be... then they all count as separate tests and if you order the decorators correctly, some combinations can be marked as skipped under certain conditions. Then when you do `python -m unittest -v` it can say "skipped this particular combination of arguments because CUDA 10.2 is dodgy" or whatever. I've done exactly that kind of thing with jUnit, I'm sure it's possible with python unittest too... if not  with decorators then [with subTest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) and skipTest.

jaycedowell: For now I've gone with a conditional expected failure approach:  if the CUDA version seems problematic then the relevant tests are marked as expected failures.  That seems to be ok if the tests actually pass.

league: I tried a little proof of concept of `unittest.TestCase.subTest` in `test_serialize` (a non-GPU test, so it could run through CI). This particular diff is fairly satisfying… it will report the 3 runs independently (with the value of the parameter) if any one of them skips or fails, though if all succeed it's counted as one test. However, `subTest` is new in python 3.4, so I assume this wouldn't pass through the CI with 2.7. Maybe another time.

```diff
     def test_serialize_with_name_no_ringlets(self):
-        self.run_test_serialize_with_name_no_ringlets()
-        self.run_test_serialize_with_name_no_ringlets(gulp_nframe_inc=1)
-        self.run_test_serialize_with_name_no_ringlets(gulp_nframe_inc=3)
+        for g in (0,1,3):
+            with self.subTest(gulp_nframe_inc = g):
+                self.run_test_serialize_with_name_no_ringlets(gulp_nframe_inc=g)
```

jaycedowell: This might be a viable WAR:  https://github.com/pygame/pygame/pull/2092

PR #162, opened 2022-03-09
Nix build using flake

league:
This is a start on specifying a “flake” for building bifrost with Nix. A flake is a format that pins all dependencies to support reproducible builds. This initial version includes github workflow for building with different releases of the nixpkgs tree (containing compilers, python dependencies, etc.) and different versions of python, then running the tests on all those configurations. (This builds on autoconf.) More background: [nixos.org](https://nixos.org/), [Flakes on Nix wiki](https://nixos.wiki/wiki/Flakes), [Flakes tutorial from tweag.io](https://www.tweag.io/blog/2020-05-25-flakes/)

It doesn't yet support GPU builds, though I've gotten it _mostly_ to work. It's a little tricky to integrate because the `libcuda.so.1` must come from the host platform, so it agrees with the kernel version and GPU hardware. If the host is itself NixOS it's manageable, but when Nix is being used on top of another platform (e.g. Ubuntu) we can only build against stubs, and then sub in the real libcuda later. For similar reasons, auto-detecting the right GPU architecture during the build seems to be problematic. As long as GPU architecture is an _input_ to the build, it gets hashed into the package signature, but we can't ask what GPU architecture is “from the inside.” Same story as for `builtins.currentSystem` for the overall architecture/OS tag. [Some hints about libcuda on Nix](https://github.com/NixOS/nixpkgs/issues/11390)

All this should be solvable (and hopefully useful). I'd like to continue to work on it and tweak it here... so marking this as a draft.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -71,14 +71,13 @@ jobs:
 
       - name: Deploy documentation to gh-pages
         uses: JamesIves/github-pages-deploy-action@v4.2.5
-        # Change github.ref below to master after #162 merged.  Might
-        # only deploy if changes in doc/ folder?  (But some doc pages
-        # generated from code, so maybe should reflect code changes.)
+        # Maybe only deploy if changes in doc/ folder?  (But some doc
+        # pages generated from code, so should reflect code changes?)
         if: |
           matrix.os == 'ubuntu-latest'
           && matrix.nixpkgs == 'pinned'
           && github.event_name == 'push'
-          && github.ref == 'refs/heads/nix-flake'
+          && github.ref == 'refs/heads/master'
         with:
           branch: gh-pages
           folder: result-doc/share/doc/bifrost/html
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,81 @@
+---
+name: "Nix Build"
+"on": [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+        nixpkgs:
+          - pinned
+          - release-21.11
+          - release-21.05
+          - nixpkgs-unstable
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2.4.0
+
+      - name: Install nix
+        uses: cachix/install-nix-action@v16
+        with:
+          extra_nix_config: |
+            access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
+
+      - name: Link to bifrost binary cache on cachix.org
+        uses: cachix/cachix-action@v10
+        with:
+          name: bifrost
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - name: Lock the version of nixpkgs specified by the matrix
+        run: |
+          if "${{ matrix.nixpkgs }}" != "pinned"; then
+            nix flake lock --override-input nixpkgs \
+              "nixpkgs/${{ matrix.nixpkgs }}"
+          fi
+          nix flake metadata
+
+      - name: Check for well-formed flake
+        run: |
+          nix flake check
+
+      - name: Build for default python3
+        run: |
+          nix build --print-build-logs .#bifrost-py3
+
+      - name: Build with enable-debug for python3, if pinned nixpkgs
+        if: matrix.nixpkgs == 'pinned'
+        run: |
+          nix build --print-build-logs .#bifrost-py3-debug
+
+      - name: Build for python38, unless nixpkgs-unstable
+        # For unstable, 3.8 packages are no longer in the build cache
+        if: matrix.nixpkgs != 'nixpkgs-unstable'
+        run: |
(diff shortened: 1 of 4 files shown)


jaycedowell:
With regards to auto-detecting the GPU arch., would something like building all available archs. be a workable solution?  It would take forever to compile but it should be generic.

league:
I'd like to find a way for telemetry to quietly disable itself if it turns out that $HOME doesn't exist or is not writable. It prevents me from ever doing `import bifrost` in a nix build environment.

```
bifrost-doc>   File "/nix/store/mdvjrgc0m3lhcz75hvcpz9lgnis69wbv-python3-3.9.6-env/lib/python3.9/site-packages/bifrost/telemetry.py", line 52, in <module>
bifrost-doc>     os.mkdir(os.path.join(os.path.expanduser('~'), '.bifrost'))
bifrost-doc> FileNotFoundError: [Errno 2] No such file or directory: '/homeless-shelter/.bifrost'
```

I was toying with generating the documentation from nix (and maybe restoring the auto-update of gh-pages as a workflow action). To generate the python API docs, it needs a python from which it can import bifrost... which can be arranged but the telemetry bombs in that environment.

I guess that whole preamble could be wrapped in a try/catch and on FileNotFound set `TELEMETRY_ACTIVE = False` and `_INSTALL_KEY` to a fresh uuid. It would still fail if somebody called enable or disable because you can't unlink or write to `_ACTIVE_KEY`, but that's probably okay… at least you should be able to import telemetry.

jaycedowell:
Now that #157 has been merged we should probably close this PR and open a new one that targets master.

league:
Yep. I may just rebase/squash onto new master... the timeline is pretty confusing with occasional merge commits from autoconf... but ultimately it should just add 3 files.

league:
I may merge the nix stuff soon, but I feel like it deserves a small section of the README or manual... maybe before next release. I'll add a CHANGELOG entry as a placeholder. Next time you're on qblocks, you might try a quick nix setup like this:

```
# Install nix
  wget https://nixos.org/nix/install
  sh install
# Log out and back in, or source the script given at the end of the install to set up current shell
# Optional: install our binary cache, maybe another useful tool or so
  nix-env -i cachix ripgrep
  cachix use bifrost
# Some basic nix config: (unfree for nvidia)
  mkdir -p ~/.config/{nix,nixpkgs}
  echo "experimental-features = nix-command flakes" > ~/.config/nix/nix.conf
  echo "{ allowUnfree = true; }" > ~/.config/nixpkgs/config.nix
```

Even before a git clone, you should be able to do stuff like the following. (If it needs to configure and build it will, but if you hit exactly the configuration that's in cachix, it will just download. Cachix is usually populated by the CI.)

- Run ctypesgen tool (after merge, won't need `/nix-flake` branch specifier)
  ```
  $ nix run github:ledatelescope/bifrost/nix-flake#ctypesgen-py3 --
  ERROR: No header files specified
  ```
- Load a python with basic (CPU-only) bifrost
  ```
  $ nix run github:ledatelescope/bifrost/nix-flake#python3-bifrost  
  Python 3.9.6 (default, Jun 28 2021, 08:57:49) 
  [GCC 10.3.0] on linux
  Type "help", "copyright", "credits" or "license" for more information.
  >>> import bifrost
  >>> bifrost.version.__version__
  '0.10.0'
  ```
- Same, but python 3.8:
  ```
  $ nix run github:ledatelescope/bifrost/nix-flake#python38-bifrost
  Python 3.8.12 (default, Aug 30 2021, 16:42:10) 
  [GCC 10.3.0] on linux
  Type "help", "copyright", "credits" or "license" for more information.
  >>> import bifrost
  ```
- Check on its config (note the double dash needed to separate the `nix run` options from the python options).
  ```
  $ nix run github:ledatelescope/bifrost/nix-flake#python3-bifrost -- -m bifrost.version --config
  bifrost 0.10.0
  Copyright (c) 2016-2020, The Bifrost Authors. All rights reserved.
  Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
  License: BSD 3-Clause
  
  Configuration:
   Memory alignment: 4096 B
   OpenMP support: yes
   NUMA support no
   Hardware locality support: no
   Mellanox messaging accelerator (VMA) support: no
   Logging directory: /dev/shm/bifrost
   Debugging: no
   CUDA support: no
  ```
- The debug-enabled version is pre-configured in the flake, just named `python3-bifrost-debug`:
  ```
  $ nix run github:ledatelescope/bifrost/nix-flake#python3-bifrost-debug -- -m bifrost.version --config
  bifrost 0.10.0
  [etc...]
   Logging directory: /dev/shm/bifrost
   Debugging: yes
   CUDA support: no
  ```
- CUDA versions available too, it's just that it will have to download the pinned cuda toolkit from nvidia (won't use the one from the underlying system) and that can take a surprising amount of time. This one probably won't be pre-built in the cache because CUDA versions aren't built by CI.
  ```
  # Make sure we can find libcuda.so.1:
  $ export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:
  $ nix run github:ledatelescope/bifrost/nix-flake#python3-bifrost-cuda11 -- -m bifrost.version --config
  ```

PR #163, opened 2022-03-10
Support `python -m bifrost.version` command

league:
Just enables a simple command to print out version and copyright info:

```
% python -m bifrost.version
bifrost 0.9.0.dev+g2bc27cc
Copyright (c) 2016-2020, The Bifrost Authors. All rights reserved.
Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
License: BSD 3-Clause
```

I think it would be nice to offer many of the 'tool' scripts with this interface, e.g. `python -m bifrost.telemetry --enable` rather than (or I guess in addition to) a script file in `bin/`.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -35,7 +35,7 @@ import sys
 import glob
 
 # Parse version file to extract __version__ value
-bifrost_version_file = 'bifrost/version.py'
+bifrost_version_file = 'bifrost/version/__init__.py'
 try:
     with open(bifrost_version_file, 'r') as version_file:
         for line in version_file:
@@ -50,8 +50,8 @@ except IOError:
     if 'clean' in sys.argv[1:]:
         sys.exit(0)
     print("*************************************************************************")
-    print("Please run `configure` and `make` from the root of the source tree to    ")
-    print("generate version.py                                                      ")
+    print("Please run `configure` and `make` from the root of the source tree to")
+    print(f"generate {bifrost_version_file}")
     print("*************************************************************************")
     raise
 
--- a/python/setup.py
+++ b/python/setup.py
@@ -51,7 +51,7 @@ except IOError:
         sys.exit(0)
     print("*************************************************************************")
     print("Please run `configure` and `make` from the root of the source tree to")
-    print(f"generate {bifrost_version_file}")
+    print("generate", bifrost_version_file)
     print("*************************************************************************")
     raise
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 doc
 version.py
+/python/bifrost/version/__init__.py
 *_generated.py
 .log*.txt
 test/data/
--- a/python/Makefile.in
+++ b/python/Makefile.in
@@ -3,7 +3,7 @@ include ../config.mk
 
 INC_DIR = ../src
 
-BIFROST_PYTHON_VERSION_FILE  = bifrost/version.py
+BIFROST_PYTHON_VERSION_FILE  = bifrost/version/__init__.py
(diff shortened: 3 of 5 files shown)


python/setup.py
```diff
@@ -50,8 +50,8 @@
     if 'clean' in sys.argv[1:]:
         sys.exit(0)
     print("*************************************************************************")
-    print("Please run `configure` and `make` from the root of the source tree to    ")
-    print("generate version.py                                                      ")
+    print("Please run `configure` and `make` from the root of the source tree to")
+    print(f"generate {bifrost_version_file}")
```
league, line 54: Oops, I guess the f-string syntax hoses 2.7 for no particularly compelling reason. Easy to avoid, but are we ready to cut off 2.7 entirely and remove it from CI?

jaycedowell: Someday we should drop support for 2.7 but I'm not sure that I'm entirely ready to do that now.

jaycedowell:
It looks like there is also a reference to `version.py` in `bifrost/__init__.py`.

jaycedowell:
I took a look through `tools` and telemetry is the only one that I see that would be a good candidate for the `python -m ...` treatment.

league:
When you did the `python -m` treatment for LSL, you didn't encounter this message:

```
RuntimeWarning: 'BLAH.__main__' found in sys.modules after import of package 'BLAH',
but prior to execution of 'BLAH.__main__'; this may result in unpredictable behaviour
```

It looks like you still have just one module `lsl.misc.telemetry` that includes an `if __main__` block. When I tried it that way for Bifrost telemetry (or version), I got the above warning. Getting around it seemed to mean using a separate `__init__.py` for the module definitions and then `__main__.py` for the program. And I think it's precisely because telemetry is imported everywhere (and version is imported by the bifrost main module). In general, if the module is not already imported before you try to run it, having an `if __main__` block will be fine.

Or having the script live at `bifrost.tools.telemetry` but the module imported everywhere is `bifrost.telemetry`, which is less satisfying.

jaycedowell:
I didn't get that warning but I also have extensively tested my changes.

jaycedowell:
Ok, well, that's fixed in LSL.

Do you want to go ahead and expand this to cover bifrost.telemetry?  I could tackle that one if it would help.

league:
@jaycedowell: I believe I have a "git stash" where `-m bifrost.telemetry` more-or-less works, so I could probably push that to the autoconf branch. However, for the purpose of building documentation (and possibly test suites etc.) I have a minor revision of telemetry.py on commit 6ad2617c1081651a6d832b35897a6e009b00217d that would be nice to include – or just address the conflict when I'm ready to merge #162... either way I guess.

If you think 6ad2617 is fine, I could cherry-pick that onto autoconf, fix its conflict with my stashed telemetry-main, and commit.

league:
Update: I see you did something similar to telemetry in lwa-project/lsl@0d12197cfe7ef2589bbc05ce9f95c7f95594890a.

jaycedowell:
Yeah, I think that is fine.  The alternative to disabling the telemetry would be to generate a new `_INSTALL_KEY` each time and report with that.  I don't think that is necessary since this readonly condition seems like an edge case.

jaycedowell:
Thinking about your comment in #164, what about doing something like having `python -m bifrost.version` give you what it does now and then adding something like a `python -m bifrost.version --config` that gives you that, plus how the library is configured.  That might be helpful for people who are on a system where Bifrost is already installed and they might be curious how it is configured.

league:
Yes! That looks very helpful. Small friendly amendment forthcoming… haha.

jaycedowell:
Does this need anything else or is it ready to merge?

league:
Ready to go.

PR #168, opened 2022-03-26
Global copyright update

league:
Yay or nay? Figured out a sed one-liner:

```
sed -i 's/\(Copyright (c) 20[0-9]\{2\}\)\(-2020\)\?\(, .*Bifrost Authors.*\)/\1-2022\3/' **/*(.)
```

Though maybe `**/*(.)` is zsh-specific to match all regular files
recursively. Could be converted to a find command. Or could just
continue doing this whenever a file is touched, I guess.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2016, The Bifrost Authors. All rights reserved.
+Copyright (c) 2016-2022, The Bifrost Authors. All rights reserved.
 Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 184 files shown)


jaycedowell:
The approach that I was trying to use was to update the copyright line when I made a change.  My thinking was that it would give me a quick way to see when a file was last updated.  I'm not sure how well that actually works since it requires someone to remember to actually change it as part of an update.  Plus, there is always the git history which is slightly more descriptive than a year.

league:
Yeah that's sensible, I'd reach for `git log FILENAME` or git blame/annotate. Some time back, my emacs would find and update copyright dates on save, which would provide some automation for your approach, but I don't have that set up anymore.

I thought of this because some big project I was navigating recently would sometimes do commits labeled "tree-wide" where they'd run a sed or other filter across everything. Keep or close, it doesn't matter to me!

PR #170, opened 2022-03-28
In Socket.hpp, prevent warnings about sock_type comparison

league:
It looks like we're trying to use `enum sock_type` to impose a more
limited type on the `int` specified by `::socket`. But it wasn't all
the way supported, because the constructor took a general
`int` (doesn't seem to be used that way within bifrost) and there were
several comparisons between `SOCK_STREAM` (respectively `SOCK_DGRAM`)
integer constants and enum-enabled `BF_SOCK_STREAM` (`BF_SOCK_DGRAM`).

So this takes that goal further, using the enum version only. An
alternative though (in case some other client code is using a general
integer for other socket types) is just to eliminate `enum sock_type`
and (if necessary) throw an exception if passed a `SOCK_*` constant
other than those we support.

The warnings were...

```
In file included from address.cpp:30:
Socket.hpp: In member function 'void
   Socket::bind(sockaddr_storage, int)':
Socket.hpp:584:12: warning: comparison between
   'enum Socket::sock_type' and 'enum
   __socket_type' [8;;
   https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wenum-compare
   -Wenum-compare8;;]
  584 |  if( _type == SOCK_STREAM ) {
      |            ^
Socket.hpp:584:12: warning: comparison between
   types 'Socket::sock_type' and '__socket_type' [
   8;;
   https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wsign-compare
   -Wsign-compare8;;]
Socket.hpp: In member function 'void
   Socket::connect(sockaddr_storage)':
Socket.hpp:596:26: warning: comparison between
   'enum Socket::sock_type' and 'enum
   __socket_type' [8;;
   https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wenum-compare
   -Wenum-compare8;;]
  596 |                    _type == SOCK_DGRAM &&
      |                          ^
Socket.hpp:596:26: warning: comparison between
   types 'Socket::sock_type' and '__socket_type' [
   8;;
   https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wsign-compare
   -Wsign-compare8;;]
```

### DIFF
--- a/src/Socket.hpp
+++ b/src/Socket.hpp
@@ -246,9 +246,7 @@ class Socket {
 	// Not copy-assignable
 	Socket(Socket const& );
 	Socket& operator=(Socket const& );
-#if __cplusplus >= 201103L
 	inline void replace(Socket& s);
-#endif
 	// Manage an existing socket descriptor
 	// Note: Accessible only via the named constructor Socket::manage
 	struct ManageTag {};
@@ -287,11 +285,9 @@ public:
 	}
 	
 	virtual ~Socket() { this->close(); }
-#if __cplusplus >= 201103L
 	// Move semantics
 	inline Socket(Socket&& s)                           { this->replace(s); }
 	inline Socket& operator=(Socket&& s) { this->close(); this->replace(s); return *this; }
-#endif
 	inline void swap(Socket& s);
 	// Address generator
 	// Note: Supports UNIX paths, IPv4 and IPv6 addrs, interfaces and hostnames
@@ -925,7 +921,6 @@ int Socket::addr_from_interface(const char* ifname,
 	::freeifaddrs(ifaddr);
 	return found;
 }
-#if __cplusplus >= 201103L
 void Socket::replace(Socket& s) {
 	_fd          = s._fd; s._fd = -1;
 	_type        = std::move(s._type);
@@ -936,7 +931,6 @@ void Socket::replace(Socket& s) {
 	_msgs        = std::move(s._msgs);
 	_iovecs      = std::move(s._iovecs);
 }
-#endif
 void Socket::swap(Socket& s) {
 	std::swap(_fd,          s._fd);
 	std::swap(_type,        s._type);
--- a/src/Socket.hpp
+++ b/src/Socket.hpp
@@ -275,16 +275,15 @@ public:
 #endif
 		DEFAULT_MAX_CONN_QUEUE = 128
 	};
-	enum sock_type {
-	  BF_SOCK_DGRAM  = SOCK_DGRAM,
-		BF_SOCK_STREAM = SOCK_STREAM
-	};
 	// Manage an existing socket (usually one returned by Socket::accept())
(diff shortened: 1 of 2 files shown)


jaycedowell:
I remember adding in the `BF_SOCK_...` enum as part of the MacOS support work for `autoconf` but I don't remember what caused me to do it.  If this isn't really needed I would be in favor of converting `::socket` back to `int` for consistency and throwing an exception for unsupported values.

league:
Yeah, here's a simplification by removing `enum sock_type`. I think this should be fine. It shouldn't throw any warnings, and removes the redundant constants for `BF_SOCK_*`. If one needs the `SOCK_*` constants from Python code, they are in `socket` module.

src/Socket.hpp
```diff
@@ -275,16 +275,15 @@ class Socket {
...
 	// TODO: With C++11 this could return by value (moved), which would be nicer
 	inline static Socket* manage(int fd) { return new Socket(fd, ManageTag()); }
-	inline explicit       Socket(/*sock_type*/int type=SOCK_DGRAM)
-		: _fd(-1), _type((sock_type)type), _family(AF_UNSPEC),
-		  _mode(Socket::MODE_CLOSED) {}
+	inline explicit       Socket(int type=SOCK_DGRAM)
+		: _fd(-1), _type(type), _family(AF_UNSPEC),
+		  _mode(Socket::MODE_CLOSED) {
+		BF_ASSERT_EXCEPTION(type == SOCK_DGRAM || type == SOCK_STREAM,
+												BF_STATUS_INVALID_ARGUMENT);
```
league, line 285: Hmm, indentation here is strange. I'm not accustomed to using tabs, but I let emacs retain tabs in files in this project for consistency, though I may be setting `tab-width` to 2. So then I think it inserted all tabs to get the 2nd line to line up after the parens, but doing that only works out if `tab-width` remains 2. Is "the right thing" to use tab char to line up with `BF_ASSERT` on first line, then spaces to get out to the parens? (This is why I don't use tabs, but I guess if I convince emacs to do the right thing it won't matter much.)

jaycedowell: Yeah, I run across strange indentation sometimes as well.  Maybe we should consider making to a move to all spaces, no tabs as some point.

league:
This all looks like an improvement now… pleased to be rid of those ifdefs too. Fine to ship, IMO.

Re whitespace, I'd def be happy with spaces-only. Some projects use an "[editor config](https://editorconfig.org/)" to help promote consistency. Less obtrusive than a commit hook and less opinionated than a full-on reformatter. I never bothered to get emacs to read and adapt to these, but I would if we use it. Also, a tree-wide untabify (or any other reformat) would best be done when there are very few unmerged branches lying around.

jaycedowell:
We should definitely untabify but I'm happy to kick that can down the road.

PR #174, opened 2022-04-07
Self-hosted GPU testing, plus make `bifrost.ndarray.copy` more robust

jaycedowell:
This PR has two parts:
 1) Add a new "build and test" workflow target that uses the new Bifrost test machine to run GPU testing.
 2) Make `bifrost.ndarray.copy` more robust with arrays that are not C contiguous, i.e., arrays that have been transposed with a stride manipulation.

(1) is largely self-explanatory but (2) has a few implementation details.  For (2) there is now a check for this particular non-contiguous condition in `copy` and Bifrost tries to do the right thing:
 * For arrays that can be accessed through the system space the Bifrost array is converted to a numpy array, copied to a new numpy array to make it C contiguous, copied back to a Bifrost array, and then copied into a Bifrost array in the correct space.
  * For arrays that can be access through the CUDA space the stride reordering is determined and `bfTranpose` is used to reorder the data.  This reorded data goes into  an intermediate Bifrost array before being copied to a Bifrost array in the correct space.

This implementation seems to work but it is copy heavy with all of the intermediates.  There is also a new test in `test_ndarray` that checks this particular condition for system-accessible arrays.


jaycedowell:
I thought this was working but I'm adding additional tests to `test_ndarray` and now it doesn't look like it.

jaycedowell:
Everything seems to checkout now with 5990366.  The only thing would be if there was a more efficient way to implement `copy`.

league:
It looks like the skip-duplicate-actions is working out. They still show up as jobs, but I guess it's easy enough to click onto the real ones. (From what I was reading, it's hard for to eliminate the job entirely as long as you want to run for push-to-branch by maintainers but also new-pull-request from outsiders.)

I was thinking it might also help to simplify the build matrix, to be more judicious or selective about what each configuration "brings to the table" rather than just multiply everything by everything. We want:

- Linux and Mac
- Python 2.7 and a handful of 3.x
- No CUDA vs CUDA 10.x vs CUDA 11.x
- Compiler that supports C++ 11 vs 14 vs 17
- With/without map cache, numa, vma, and any other `--enable` flags or `#ifdef`s
- With nix or without

But **every** possible combination is probably not productive. If we push something that works with python 3.8 but causes an issue with 3.6, it's helpful to know that – but the issue is **probably** consistent across most of the other permutations.

jaycedowell:
If I had to guess what combinations would be useful:
 * Py2.7 + ubuntu
 * Py3.6 + ubuntu
 * Py3.8 + ubuntu
 * Py3.8 + macos
 * Py3.8 + ubuntu + CUDA10
 * Py3.8 + ubuntu + CUDA11
 * Py3.8 + ubuntu + CUDA11 + "options"

jaycedowell:
I don't know that I've come up with a good way to roll different CUDA versions in with the self hosted runner.  The best I've been able to come up with is to alternate the container that is used between self hosted runs.  It's like testing multiple versions of CUDA but with unknown combinations of other parameters.

league:
> different CUDA versions in with the self hosted runner

Might actually be something that nix can help with. Through nixpkgs, I can install cuda-X.Y and build with it – even on my GPU-less thinkpad. You just can't query the hardware (so configure needs its `--with-gpu-archs`). Hypothetically one could even do those builds on GitHub CI machines. (The built artifacts won't run without GPU however.) I haven't tried doing GPU builds on GitHub yet mainly because the download from nvidia can take forever, it's a many-GB package for each one. The workflow feature to cache stuff from one build to the next could help, but I'm not sure what the size limitations are.

But on self-hosted server, maybe it would work. Can you describe more about how self-hosted works? It's building on a standard ubuntu docker image but with GPU pass-through? Can we select which image to use, and is that per-build or per-repo? I can create a branch to try out some small experiments with self-hosted.

jaycedowell:
Yes, it's based on a standard Ubuntu image with CUDA 11 installed.  It's setup as an ephemeral runner so you get only one workflow/test per container launch.  I have cron job that makes sure there is always a container running.  Everything that makes it work is on the test machine in `~jdowell/selfhosted/`.  The biggest thing that I think it is missing right now, apart from the possibility of different CUDA versions, is limiting its access to only a single GPU.

league:
> based on a standard Ubuntu image with CUDA 11 installed

I see it now. So this means that installation of dependencies can be sort of partitioned between those specified in this Dockerfile (which would transcend different builds) and those in the `main.yml` (which are repeated with each build).

And further, if we wanted any additional data/packages to be cached or imported locally, we could do that via COPY or even VOLUME in the Dockerfile.

jaycedowell:
That's right.  The Dockerfile gives us the base system+CUDA that is ~constant between tests and then main.yml provides the test environment (at least the Python portion/dependencies) as well as the tests themselves.  I will say that I'm a little nervous about using too much COPY or VOLUME to expose things to Docker.  This is partially because of wanting to isolate the container as much as possible and partially because I had some problems using long-lived containers running multiple tests when I was first setting this up.

jaycedowell:
It'd be really nice if we could get the codecov report off the self hosted runner but I have yet to figure out how to do that.

league:
Definitely, we want to see the coverage reported by a GPU runner. I'll toy with that a little today.

league:
It looks like it **is** uploading, but from multiple platforms so the self-hosted one probably doesn't "win". I'll try some stuff, but apologies if it revisits what you already figured out. :)

league:
Ah, now I see how `env_vars` works and where to find the reports per `OS`: https://codecov.io/gh/ledatelescope/bifrost/commit/35d9f702414de9feff2f234a22267f25d90f069c/build

So it's getting all three, I'm just not sure how/whether it's merging them together for the front-page numbers that it shows.

jaycedowell:
I had started out with only trying to have one coverage report from self-hosted and Py3.8 but every time I did that I ended up with some kind of "missing commit"-style message from codecov.  That's why I switched to trying multiple reports.  If multiple reports is the only answer then maybe we need a codecov YAML file so that we can set the [notify after_n_builds option](https://docs.codecov.com/docs/notifications).  Maybe that would force some kind of aggregation.

league:
I see the error now. Pretty non-specific, but there's a good chance it *does* have to do with the source code paths. I noticed in the XML-ish files I downloaded that paths were pretty different:

Ubuntu:
```
<class name="fdmt.py" filename="/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/bifrost/fdmt.py" complexity="0" line-rate="0.5" branch-rate="0">
```

Mac:
```
<class name="fdmt.py" filename="/Users/runner/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/bifrost/fdmt.py" complexity="0" line-rate="0.5" branch-rate="0">
```

Self-hosted:
```
<class name="fdmt.py" filename="/home/docker/actions-runner/_work/_tool/Python/3.8.12/x64/lib/python3.8/site-packages/bifrost/fdmt.py" complexity="0" line-rate="1" branch-rate="0">
```

They're all different, but maybe it knows how to deal with the first two. Will try their advice about fixing up the paths in the codecov yml.

jaycedowell:
That worked!

PR #176, opened 2022-04-22
Pass `sockaddr_storage` by reference

jaycedowell:
This PR updates `Socket.hpp` to pass `sockaddr_storage` by reference to `bind()`, `connect()`, `address_string()`, and `discover_mtu()`.  There are also a few small cleanups with regards to unused variables and some non-printing characters.


league:
Since it passed tests, If you're interested in also moving socket implementations into a .cpp, that's in the branch `socket-deinline` and could be merged into this PR. Or not, but I'm a little suspicious of this much inline code, I guess. I'll make another inline review comment about a warning I get on Mac with gcc.

league:
Oh, never mind inline comments, I just added the fixes to `socket-deinline`, with detailed commit messages. If you decide not to de-inline the socket header, then I can apply those small patches to the same functions in the .hpp on `socket-reference`.

jaycedowell:
@league I think I'm happy with this now that `Socket.cpp` is a thing.

Issue #156, opened 2021-11-12
duplicate functionality?


jaycedowell:
@telegraphic has noted that both `bifrost.block` and `bifrost.pipeline` provide the same functionality.  There are also several other places in Bifrost where this seems to happen:
 * DataType.py vs. dtype.py
 * ring.py vs. ring2.py
 * block.py vs. pipeline.py
We should look through these pairs to (1) see if they are indeed duplicating functionality and (2) if so, which one gets eliminated.  @telegraphic has already voted to remove block.py.

jaycedowell:
For a FYI on ring.py vs. ring2.py: ring.py is currently in use for the tutorials.  I also use it extensively in my code that uses the lower level Python API.

PR #178, opened 2022-04-29
Code Pruning

jaycedowell:
This PR cleans up the code base by removing unused and commented out code.  It may also be an appropriate place to address #156.


league:
I'm looking into the drop in coverage on ring.py. Would to complete/resolve the ring → ring2 replacement/merger with this branch?

jaycedowell:
I've given up on removing `bifrost.block` since there seems to be a few knock on consequences to that.  `bifrost.ring` vs `bifrost.ring2` might be a similar thing but I think we need to sort out #179 before we try that.

jaycedowell:
I'm starting to pick this up again.  I wonder if this kind of cleanup makes sense to do as part of the `no_python2` branch.

jaycedowell:
This project has migrated over to #216 and the `no_python2` branch.   `apr2022-prune` should be removed once #216 is merged.

PR #179, opened 2022-05-06
Merge tutorials into this repo?

league:
Would make for easier CI on them. They serve both as documentation and test. Making this into a PR so it may be easier to navigate the commits involved, etc.

This branch was **rewritten** from the bifrost_tutorial repo. All dates and authorships should be the same, but we're pretending that everything was done in a `tutorial/` subdir all along.


league:
Just don't squash-merge, then the carefully-constructed lies in the history would be lost. :)

jaycedowell:
Just to play devil's advocate:  what about using https://github.community/t/triggering-by-other-repository/16163/12 to trigger testing the tutorial when master is updated?

jaycedowell:
I'm actually leaning towards moving the tutorials into the main repo since I think it will make things easier for new users to get up and running.  Plus, it could be helpful from a testing perspective to have more examples of how Bifrost is used in additional real world-like contexts.

I threw a poll on the slack channel to see if anyone else has an opinion.

jaycedowell:
Eh, it doesn't seem to do anything to the coverage.

jaycedowell:
For tutorial 6 do we want to update that to point to the `ibverb-support` branch here on ledatelescope?

jaycedowell:
hwloc shouldn't be required for these tests.  What if you throw a `--disable-hwloc` to configure?

league:
Yes, went through with `--disable-hwloc` and also reduce `bifrost.quantize.quantize` to just `bifrost.quantize`.

jaycedowell:
Excellent.  Now I just need to fix the testing on cetus.

jaycedowell:
Things should start running now.

jaycedowell:
Personally, I would like to minimize how many external/supplementation things that we need to maintain.  I'm be happy with the AWS stuff being deleted.  Docker is harder but I'm not seeing a lot of added value there now that building Bifrost is a lot easier.  So maybe "yes, delete" for AWS and "yes, barebones" for Docker?

league:
I think I agree. The value of Amazon was only to try it out from a machine with no GPU. And I just noticed that you updated the Dockerfile in eadcf24d45517b0a7365b255bb5300b810739584 so it should be relatively easy to run and push to dockerhub once in a while. Keeping AWS images updated would be more work than that.

league:
I'll try rerun of tutorial Dockerfile on cetus.

league:
I think I'm going to merge tutorial stuff shortly.

league:
I think that one out-of-memory report on self-hosted 3.6 could be because I was simultaneously messing with a docker build. I think the Dockerfile works again. I could run jupyter through it and it had Bifrost installed and could run anything non-GPU. When running something using GPU it hanged, but I guess because both GPUs are being used by the runners. (Or maybe I need to do something more than just `--runtime=nvidia`.)

jaycedowell:
I've been doing some testing on the north arm pipelines and there are a couple of those running.  That uses some GPU memory.

PR #181, opened 2022-05-24
cufft c2r issue still persists with CUDA 11.7

### LINKED ISSUE #180: Fail tests test_fft.TestFFT on Ubuntu 20.04
Hello,
I'm starting to use bifrost and I'm happily starting with the tutorials.
However, just out of curiosity, to check my install, I run `make test`.
Many passed but `FAILED (failures=3, skipped=4)`

FAIL: test_c2r_1D (test_fft.TestFFT)
  File "/home/cedric/tmp/bifrost/test/test_fft.py", line 206, in test_c2r_1D
    self.run_test_c2r(self.shape1D, [0])
  File "/home/cedric/tmp/bifrost/test/test_fft.py", line 148, in run_test_c2r
    self.run_test_c2r_impl(shape, axes)
  File "/home/cedric/tmp/bifrost/test/test_fft.py", line 141, in run_test_c2r_impl
    compare(odata.copy('system'), known_result)
  File "/home/cedric/tmp/bifrost/test/test_fft.py", line 51, in compare
    np.testing.assert_allclose(result, gold, rtol=RTOL, atol=MTOL * absmean)
  File "/home/cedric/anaconda3/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 1528, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/cedric/anaconda3/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 842, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=0.1, atol=0.00462357
Mismatched elements: 1 / 16777216 (5.96e-06%)
Max absolute difference: 0.01639435
Max relative difference: 1.93911746

FAIL: test_c2r_2D (test_fft.TestFFT)
AssertionError: 
Not equal to tolerance rtol=0.1, atol=0.00231149
Mismatched elements: 4186048 / 4194304 (99.8%)
Max absolute difference: 492620.22392237
Max relative difference: 39550830.05759069

FAIL: test_c2r_3D (test_fft.TestFFT)
AssertionError: 
Not equal to tolerance rtol=0.1, atol=0.00163087
Mismatched elements: 2080441 / 2097152 (99.2%)
Max absolute difference: 88917.37869481
Max relative difference: 6823521.1182115

This was tested on Ubuntu 20.04.4 LTS with Python 3.8.8.

Tell me if I can help.
Regards,
Cedric

league:
“Fixes” is a strong word, but I suppose this can close #180. I'm just updating the warnings and exclusion of the c2r tests for CUDA 11.7, which were reported therein. @jaycedowell, I guess it would need an autoconf rerun before merge.

### DIFF
--- a/configure
+++ b/configure
@@ -26801,7 +26801,7 @@ fi
 
 if test x$HAVE_CUDA = x1
 then :
-  cufft_c2r_modifies=$( echo $CUDA_VERSION | ${GREP} -e "10\.[1-2]" -e "11\.[0-6]" )
+  cufft_c2r_modifies=$( echo $CUDA_VERSION | ${GREP} -e "10\.[1-2]" -e "11\.[0-7]" )
        if test "$cufft_c2r_modifies" != ""; then
          echo ""
          { printf "%s\n" "$as_me:${as_lineno-$LINENO}: WARNING: This version of cuFFT may have unexpected behavior for complex-to-real transforms" >&5
--- a/configure.ac
+++ b/configure.ac
@@ -346,7 +346,7 @@ AC_OUTPUT
 #
 
 AS_IF([test x$HAVE_CUDA = x1],
-      [cufft_c2r_modifies=$( echo $CUDA_VERSION | ${GREP} -e "10\.[[1-2]]" -e "11\.[[0-6]]" )
+      [cufft_c2r_modifies=$( echo $CUDA_VERSION | ${GREP} -e "10\.[[1-2]]" -e "11\.[[0-7]]" )
        if test "$cufft_c2r_modifies" != ""; then
          echo ""
          AC_MSG_WARN(This version of cuFFT may have unexpected behavior for complex-to-real transforms)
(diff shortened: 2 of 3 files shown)


league:
Lol that configure update was a quick response, are you a bot? Add 34957 to 70764.

jaycedowell:
My understanding of this problem is that c2r transforms in CUDA 10.1+ are allowed to change the input array.  That causes problems for Bifrost when `fftshift=True` since the transform rearranges the input while Bifrost is trying to rearrange the input.  It would be great if we could actually fix this problem.  "Fix" in this case could be either:
 1) we don't support c2r with `fftshift=True` or
 2) we catch this case and pre-shift the inputs before calling the transform.

I would prefer (2) but I'd have to think about how to do that.  Only changing the `bifrost.fft` module leaves people in C/C++ out of luck.  Changing `fft.cu` is a better option but more involved.  Before starting down either road we should verify that `fftshift=True` is the only problem and that it isn't a general c2r thing.

jaycedowell:
I've started a new `fft-c2r-fix` branch to fix the underlying issue.  My goal is to catch c2r transforms on CUDA 10.1+ and  shift the input data _before_ the cuFFT call rather than relying on the `pre_fftshift` function in the callback _during_ the cuFFT call.

PR #183, opened 2022-07-08
Support for CUDA graphs

jaycedowell:
This PR adds support for using CUDA graphs inside a Bifrost block using a new `bifrost.device.Graph` context manager.  `Graph` is used as:
```
import bifrost as bf
graph = bf.device.Graph()
for i in range(100):
    with graph:
        if not graph.created:
            <some GPU stuff here>
    bf.device.stream_synchronize()
```
For the loop over `i`:
 * `i=0`: Entering the `graph` context manager does nothing to allow for `bifrost.map` and other things that need initialization to run without interacting with the graph creation.
 * `i=1`:   Entering the `graph` context manager launches a `cudaStreamBeginCapture` call.  When `graph.__exit__()` is called when leaving the context the stream capture ends and an executable CUDA graph is created.
 * `i>=2`: Subsequent iterations over `i` then jump over the block body with the `if` statement and the the context manager executes the CUDA graph at `__exit__()`.

This should help reduce execution overheads on subsequent calls of a loop body.  The only caveat I've found is that calls to a stream synch. are not allowed inside a graph.  This means that the standard `bifrost.ndarray.copy_array` function cannot be used when the the two arrays are in different memory spaces.  To work around that `Graph` has a `copy_array` method which does the same thing as its `bifrost.ndarray` counterpart but does not try to synchronize the stream.  This is part of the reason for the call to `stream_synchronize()` after exiting the `graph` context.

This PR should be merged after #167.


jaycedowell:
I think it works but I'm not super happy with the whole `raise bf.device.GraphCreatedError` part.  I'm open to suggestions of how to restructure things.

jaycedowell:
I might be happy with [000fc53](https://github.com/ledatelescope/bifrost/pull/183/commits/000fc53b99de5bf3250ad9f6dc02a1df830551c6).  I'll update the description.

jaycedowell:
Maybe it's not ready with the latest set of changes...

jaycedowell:
Even though [e359f33](https://github.com/ledatelescope/bifrost/pull/183/commits/e359f337d031b210e4896a99492b03ac9efac76e) seems like it will work it has problems.  You can still end up in a situation where the host callback data structures are overwritten by a subsequent `bifrost.fft.Fft` call if the same object is used.  Think forward transform followed by an inverse transform.  There isn't a synch.-less way I can think of to ensure that all of the first call has been processed before the second call is scheduled.

Graphs are starting to look like the wrong solution to the problem I was trying to solve.

jaycedowell:
[cuFFTDx](https://docs.nvidia.com/cuda/cufftdx/index.html) might make a difference for this.

PR #192, opened 2022-12-06
diable -> disable typo

jack-h:
If you think this is a sign I'm fighting with merging lwa352 into the new build system....you would be correct.


league:
Thanks for the fix! The CI showed 3 spurious failures but of course it's not this commit causing them. :)

I'm working on 2 of those 3 issues right now, related to `<pyconfig.h>` on self-hosted and installing 3.6 on ubuntu-latest. So I can hopefully cherry-pick solutions onto this branch shortly.

The 3rd seems to be a legit test failure in `test_read_drx` on self-hosted 3.6. Unclear (to me) what went wrong there.

PR #194, opened 2022-12-15
PSRDADA block updates

telegraphic:
These updates are from https://github.com/Molonglo/bifrost, which is used for UTMOST-2D. 

An alternative to this merge is to move PSRDADA functionality into a plugin.


python/bifrost/blocks/psrdada.py
```diff
@@ -27,14 +27,16 @@
...
 from bifrost.psrdada import Hdu
 from bifrost.libbifrost import _bf, _check
 import bifrost.ndarray
 
-from bifrost import telemetry
-telemetry.track_module()
+import numpy as np
+from datetime import datetime
+from copy import deepcopy
+import os
```
jaycedowell, line 39: We should make sure every module has a telemetry call.

python/bifrost/psrdada.py
```diff
@@ -119,12 +123,12 @@ def __next__(self):
         else:
             del block
             self.reset()
-            raise EndOfDataStop('IpcBufBlock empty')
+            raise StopIteration()
```
jaycedowell, line 126: I thought there was some problem with `StopIteration` and that's why we moved to `EndOfDataStop`.  Maybe I'm thinking of something else?

python/bifrost/blocks/psrdada.py
```diff
@@ -129,7 +147,174 @@ def on_data(self, reader, ospans):
...
+
+    # update parameters from bifrost tensor
+    if '_tensor' in hdr_dict.keys():
+        print(hdr_dict['_tensor'])
+        dtype = hdr_dict['_tensor']['dtype']
+        dtype_vals = {
+            'cf32': { 'NBIT': '32', 'NDIM': '2' },
+            'f32': { 'NBIT': '32', 'NDIM': '1' },
+            'ci8': { 'NBIT': '8', 'NDIM': '2' },
+            'i8': { 'NBIT': '8', 'NDIM': '1' } }
```
jaycedowell, line 193: Are additional data types expected/allowed?

python/bifrost/blocks/psrdada.py
```diff
@@ -129,7 +147,174 @@ def on_data(self, reader, ospans):
...
+                          int(hdr_dict['NANT']) * \
+                          int(hdr_dict['NPOL']) * \
+                          int(hdr_dict['NBIT']) * \
+                          int(hdr_dict['NDIM']) * \
+                          int(hdr_dict['NCHAN'])
+
+        resolution = (bits_per_sample * fine_time) / 8
+        hdr_dict['RESOLUTION'] = resolution
+   
+        bytes_per_second = int((bits_per_sample / 8) / tsamp)
```
jaycedowell, line 235: // 8 then we can drop the `int()`

ledatelescope:
Tensor core arithmetic also accelerates 16 bit type data.

python/bifrost/psrdada.py
```diff
@@ -44,25 +44,29 @@
 import bifrost.libpsrdada_generated as _dada
 import numpy as np
 from bifrost.ndarray import _address_as_buffer
-from bifrost.libbifrost import EndOfDataStop
 
 import ctypes
 
-from bifrost import telemetry
-telemetry.track_module()
-
```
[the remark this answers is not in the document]
telegraphic: I manually copied the psrdada.py code over from the bifrost code @ Molonglo, which is on an out-of-date branch

python/bifrost/psrdada.py
telegraphic: Ugh, this is a regression bug

python/bifrost/blocks/psrdada.py
telegraphic: There isn't really a strict definition AFAIK. There is no header keyword for float/integer, but I think it's reasonable to add cf64/f64. Would need to decide between i16/ci16 and f16/cf16 though

python/bifrost/blocks/psrdada.py
```diff
@@ -129,7 +147,174 @@ def on_data(self, reader, ospans):
...
+
+    for key, val in hdr_dict.items():
+        if key not in keys_to_skip:
+            if isinstance(val, (str, float, int)):
+                s += _keyval_to_dadastr(key, val)
+    s_padding = "\x00"
+    if len(s) > hdrlen:
+        raise RuntimeError("Header is too large for HDR_SIZE! Increase hdrlen")
+    n_pad = hdrlen - len(s)
+    return s + s_padding * n_pad
```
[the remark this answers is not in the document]
telegraphic: very good chance there is :/. I guess I should write a unit test

python/bifrost/blocks/psrdada.py
jaycedowell: i16/ci16 and f64/cf64 sound good to me.  I'm not sure that bifrost works with f16/cf16, at least on the cuda side.  I should look into that.

python/bifrost/blocks/psrdada.py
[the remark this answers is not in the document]
jaycedowell: I don't think we have anything that covers psrdata now.  We probably don't even install psrdata to test the wrapping.

ledatelescope:
I raised the notion of 16-bit  to accommodate directly time-series output from 
many-bit ADCs.  i16 would do that.  Quick examination suggests that uint16 is 
recognized by CUDA7(?) but using it may not result in speed advantage over 
4-byte integer math.  Proper research required.

python/bifrost/blocks/psrdada.py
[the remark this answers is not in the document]
jaycedowell: The CI now installs PSRDADA and the Bifrost wrappers get generated.  @telegraphic do you have a .dada file I can test with?

PR #204, opened 2023-05-31
Only support CUDA 10+

jaycedowell:
This PR drops support for CUDA versions prior to 10.  This is done through a combination of catching old versions of CUDA in `configure` and dropping the `__CUDACC_VER_MAJOR__` tests that toggle for features before/at  9.

### DIFF
--- a/config/cuda.m4
+++ b/config/cuda.m4
@@ -38,34 +38,24 @@ AC_DEFUN([AX_CHECK_CUDA],
     LIBS_save="$LIBS"
     
     ac_compile='$NVCC -c $NVCCFLAGS conftest.$ac_ext >&5'
-    AC_COMPILE_IFELSE([
+    LDFLAGS="-L$CUDA_HOME/lib64 -L$CUDA_HOME/lib"
+    LIBS="$LIBS -lcuda -lcudart"
+
+    ac_link='$NVCC -o conftest$ac_exeext $NVCCFLAGS $LDFLAGS $LIBS conftest.$ac_ext >&5'
+    AC_LINK_IFELSE([
       AC_LANG_PROGRAM([[
           #include <cuda.h>
           #include <cuda_runtime.h>]],
-          [[#if __CUDACC_VER_MAJOR__ < 10
-             asdfdsfd
-            #endif]])],
-        [],
-        [AC_SUBST([HAVE_CUDA], [0])])
-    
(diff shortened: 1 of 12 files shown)


config/cuda.m4
```diff
@@ -42,7 +42,9 @@ AC_DEFUN([AX_CHECK_CUDA],
       AC_LANG_PROGRAM([[
           #include <cuda.h>
           #include <cuda_runtime.h>]],
-          [[cudaMalloc(0, 0);]])],
+          [[#if __CUDACC_VER_MAJOR__ < 10
+             asdfdsfd
```
league, line 46: This is fine because it makes the build fail for CUDA <10, but in the output we can't distinguish that as the failure reason. I was thinking a bit below this, where we extract CUDA_VERSION and print it, we could also do a shell-script if/else there, and produce output like "Checking for a working CUDA 10+ installation..." then either "yes - v12.0", "no - found v9.1", or "no - build error".

Something like below, but I'm hesitant to commit a change to `cuda.m4` because I still can't find a reliable way to re-autoconf. It always loses the Doxygen rules, and the C++ stuff looks really different, even when I'm using autoconf-2.71 like you are.

```diff
           [CUDA_VERSION=$( ${NVCC} --version | ${GREP} -Po -e "release.*," | cut -d,  -f1 | cut -d\  -f2 )
-           AC_MSG_RESULT(yes - v$CUDA_VERSION)],
-          [AC_MSG_RESULT(no)
+          CUDA_MAJOR=$( echo "${CUDA_VERSION}" | cut -d. -f1 )
+          if test "${CUDA_MAJOR}" -ge 10; then
+             AC_MSG_RESULT(yes - v$CUDA_VERSION)
+          else
+            AC_MSG_RESULT(no - found v$CUDA_VERSION)
+            AC_SUBST([HAVE_CUDA], [0])
+          fi],
+          [AC_MSG_RESULT(no - build error)
            AC_SUBST([HAVE_CUDA], [0])])
```

jaycedowell: Yes, I get a message that is similar to that.

I hadn't included them because some of existed on my Ubuntu system and I didn't want to fight with potential version conflicts at the time.  Maybe that's not really an issue.

jaycedowell:
Anything else you want to add before I merge this?

league:
Nope, go. 🛳️

PR #207, opened 2023-06-01
Fix transpose to work on single pol data, with test

dentalfloss1:
The diff is actually correct this time! Fixes transpose to work on single polarization data. Includes working with mono wav files in the tutorial. Added a test for these kinds of datasets.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,8 +23,6 @@ jobs:
         python-version: ['3.8', '3.10']
         include:
           - os: ubuntu-20.04
-            python-version: '2.7'
-          - os: ubuntu-20.04
             python-version: '3.6'
           - os: macos-latest
             python-version: '3.6'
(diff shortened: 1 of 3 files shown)


league:
The successful test speaks for itself! But I'm trying to understand what was broken about it in the first place, and I'm wondering if it can be fixed without special-casing `K==1`. Was the issue maybe because `in_inds_str` contained a reference to `k` when it's mono? Then `k` wouldn't exist until inside the loop, and in this fix, `k` is declared without a loop (because the loop would be singleton anyway). Would it be the same to produce something like:

```cpp
func_str += "enum { K = " + std::to_string(K) + " };\n";
func_str += "int k=0;\n";
func_str +=
    "in_type ivals = in(" + in_inds_str + ");\n"
    "#pragma unroll\n"
    "for( ; k<K; ++k ) {\n"
    "    out(" + out_inds_str + ") = ivals[k];\n"
    "}\n";      
```

Does that work-alike when `K==1` ? Maybe I'm confused about why it wasn't working in the first place?

league:
I was able to try what I _thought_ might unify the `K==1` case with the general case. But it doesn't work! Passing the test is the main thing, so I think this fix can be merged!

When it's passing the tests, here is an example of generated code when `K==1`:

```cpp
enum { K = 1 };
int k=0;
out_type ovals = in(k, i0, i1);
out(i0, i1) = ovals;
```

compared to when `K==2`:

```cpp
enum { K = 2 };
in_type ivals = in(i2, i1);
#pragma unroll
for( int k=0; k<K; ++k ) {
    out(k, i1, i2) = ivals[k];
}
```

Oh, one little simplification incoming in a moment…

PR #213, opened 2023-06-21
Update create a pipeline

dentalfloss1:
I updated the first part of Create-A-Pipeline in order to make the pipeline initialization a bit more readable. I also created a new section to read back the filterbank file and make a waterfall plot of the results.


jaycedowell:
Overall this LGTM.  My only comment is that there is this talk about figuring out the bpm of the song at the beginning that doesn't really seem happen.  Do we want to try to fold that in or drop that?

dentalfloss1:
That sounds like a good idea, but I'm not sure how it could be executed. Perhaps look at the spectrograph for regular bumps in signal over time? Unfortunately I think that would require a specific example. I can look into it further though
________________________________

jaycedowell:
I'm not entirely sure how you would get to beats per minute.  It could be something like summing over frequencies and then running another FFT to create a secondary spectrum.

Issue #187, opened 2022-10-17
Intermitten `bifrost.linalg` test failures


jaycedowell:
Occasionally we see test failures on the self-hosted `bifrost.linalg` suite.  Now that I'm looking for one to point to I cannot find one.

jaycedowell:
Here's one:  https://github.com/ledatelescope/bifrost/pull/167#issuecomment-1152494636

jaycedowell:
I wonder if this is somehow related to #210.

PR #215, opened 2023-07-20
Update dockerfiles to ubuntu-20.04 and autoconf

league:
The nvidia/cuda:10.2 image disappeared from the hub, so the GPU versions in master were not buildable anymore.  Also, the two that build bifrost were not yet using the configure script.

However, the full-build `.gpu` version doesn't quite work because nvidia-docker seems to provide access to GPU during run phase, but not during build phase.  Perhaps relevant:
https://stackoverflow.com/questions/59691207/docker-build-with-nvidia-runtime

The `_prereq.gpu` version builds fully and is still helpful.

Also pinging PR #92 which is outdated but relevant.


jaycedowell:
Oh look, it's our good friend the "test_matmul_aa_ci8" test failure (#187).

jaycedowell:
For the build of .gpu could we just specify all supported SMs?  And then set a reasonable default for the shared memory per SM?  I think those are the only two things that require a working GPU.

jaycedowell:
Should we turn the docker builds into CI tests?  Maybe only in master?

league:
I'm mixing another idea into this dockerfile PR, but it's minor stuff I encountered about test conditions while figuring out the prereqs in docker.

Re: building with Dockerfile.gpu, yes I think specifying to configure will work, will try. It's similar to what we need with nix because of hermetic build where we can't directly query hardware capabilities.

Re: building docker in CI: I wasn't sure whether docker-build works within docker but I guess it might. May  be worthwhile if we want to support these things. (They are still mentioned in the README, and I guess I'm still finding them convenient for testing and playing around.)

jaycedowell:
Maybe this could help with the docker stuff:  https://github.com/marketplace/actions/build-and-push-docker-images

jaycedowell:
Maybe we should go up to Jammy on this.

jaycedowell:
NB:  Update to universal ctags (#250).

jaycedowell:
The `test_matmul_ab_beamformer_kernel_large` and `test_matmul_ab_beamformer_kernel_small` failures are related to the shared memory size (16K just isn't enough).  With CUDA 12.6 supporting down to compute capability 50 I think we are safe to increase the default shared memory to 48K (what devices with cc 50 had).

Between [d951aae](https://github.com/lwa-project/bifrost/pull/215/commits/d951aaed8bd1915751f294c267c9e72de11789fc) and [85e0997](https://github.com/lwa-project/bifrost/pull/215/commits/85e09976366ee5a718e98a1e50ba6f8dd7f9c238) that just leaves us with the `test_complex_reduce` failure.

jaycedowell:
Everything looks good and I'm going to sidestep the issue of "docker builds as a CI test" for now.

PR #216, opened 2023-07-20
Ditch python2

league:
This was a branch, but I just updated it from master and started a PR to more easily see and discuss changes.

(I was about to delete some `__future__` stuff when I remembered to look for this branch.)


jaycedowell:
It might be good to pick some of the "safer" things from #178 and add them in here.

league:
I think all of the clean-ups in #178 are here now, except the ones that are more complicated:
- [ ] removing block.py
- [ ] ring vs ring2

But commits in this PR (or that previously made it to master) include these clean-ups:
- [x] remove dtype.py
- [x] pypy stuff
- [x] commented-out code in ring.py
- [x] remove test_fft.cu

…in addition to the Python2 cleanups.

jaycedowell:
I recall problems when I tried to remove block.py but I don't remember exactly what they were.   It's a shame that those logs from #178 have expired and been removed.

jaycedowell:
When I was working on [8cff5c2](https://github.com/ledatelescope/bifrost/pull/216/commits/8cff5c27be4c7d8cae70d9f9f561a42e7d435b14) I noticed that there was another Python2-ism buried in `test_scripts.py`.  There might be other things labeled as "Python2 catch" in the code, specifically related to bytes -> str -> bytes conversion that were missed.

jaycedowell:
Should we put a "needs Python 3.6+" message somewhere?  That's the only thing I can thing of that this might be missing.

tools/like_ps.py
```diff
@@ -65,12 +62,8 @@ def get_process_details(pid):
 
     data = {'user':'', 'cpu':0.0, 'mem':0.0, 'etime':'00:00', 'threads':0}
     try:
-        output = subprocess.check_output('ps o user,pcpu,pmem,etime,nlwp %i' % pid, shell=True)
-        try:
-            output = output.decode()
-        except AttributeError:
-            # Python2 catch
-            pass
+        output = subprocess.check_output(f"ps o user,pcpu,pmem,etime,nlwp {pid}", shell=True)
```
jaycedowell, line 65: Does this work if we use a:
```
output = subprocess.check_output(['ps', 'o', 'user,pcpu,pmem,etime,nlwp', str(pid)])
```
?

tools/pipeline2dot.py
```diff
@@ -65,10 +62,8 @@ def get_process_details(pid):
 
     data = {'user':'', 'cpu':0.0, 'mem':0.0, 'etime':'00:00', 'threads':0}
     try:
-        output = subprocess.check_output('ps o user,pcpu,pmem,etime,nlwp %i' % pid, shell=True)
-        if sys.version_info.major > 2 and isinstance(output, bytes):
-            # decode the output to utf-8 in python 3
-            output = output.decode("utf-8")
+        output = subprocess.check_output(f"ps o user,pcpu,pmem,etime,nlwp {pid}", shell=True)
```
jaycedowell, line 65: This would be another place to think about if we need `shell=True` or not.

configure.ac
```diff
@@ -232,7 +232,12 @@ AS_IF([test x$enable_python != xno],
               AS_IF([test x${PYTHON3} != xno],
                     [AC_SUBST([PYTHON], [$PYTHON3])])])
        AS_IF([test x${PYTHON} != xno],
-             [AC_MSG_CHECKING([whether $PYTHON as ctypesgen])
+             [AC_MSG_CHECKING([if $PYTHON is version 3.6 or later])
+              AS_IF([! ${PYTHON} -c "import sys; assert(sys.version_info >= (3,6,0))" 2>/dev/null],
+                    [AC_MSG_RESULT([no])
+                     AC_MSG_WARN([python module will not be built])],
+                    [AC_MSG_RESULT([yes])])
+              AC_MSG_CHECKING([whether $PYTHON as ctypesgen])
```
jaycedowell, line 240: We might want to make these two checks nested so that we don't look for `ctypesgen` if we don't have Python >=3.6.

jaycedowell (commented):
I fixed most of the things that I noticed but there are a couple left.

configure.ac
jaycedowell: See [ff06f45](https://github.com/ledatelescope/bifrost/pull/216/commits/ff06f4539f41f9b9d07169fe9b7af2519ffb9cb2).

tools/pipeline2dot.py
jaycedowell: See [f5b0b02](https://github.com/ledatelescope/bifrost/pull/216/commits/f5b0b02cb1cc8cd446a3f4d90cfc824276c56b64).

tools/like_ps.py
jaycedowell: See [f5b0b02](https://github.com/ledatelescope/bifrost/pull/216/commits/f5b0b02cb1cc8cd446a3f4d90cfc824276c56b64).

jaycedowell:
That's everything now.

PR #217, opened 2023-07-21
Remove the deprecated numpy dtypes "int" and "float"

jaycedowell:


jaycedowell:
It looks like this has sorta kinda already been merged in but with `numpy.int64` replacing `numpy.int` and `numpy.float64` replacing `numpy.float`.  I _think_ those are the correct sizes for native Python int's and float's so maybe this PR isn't needed?

jaycedowell:
@dentalfloss1 @league  Any opinions on this PR which would allow lower precision values to be stored in a "standard header"?  I'm currently leaning towards keeping the current behavior in `master` and not merging this one.

dentalfloss1:
I agree.

jaycedowell:
Motion carries, closing.

PR #219, opened 2023-08-18
Fix dep warning

dentalfloss1:
Deprecation warnings should be gone. Notable is the change from np.int8 to np.uint8 in the ci4 datatype. One test failed because I was testing against numpy 1.25 and one of our tutorial files still requires numba, which is incompatible with numpy 1.25


jaycedowell:
I noticed that a few of the CI tests failed so I'm rerunning them.

PR #166, opened 2022-03-22
Unified packet capture interface, take 2

jaycedowell:
This PR provides a unified packet capture interface that makes it easier to add new packet formats to Bifrost and to read packet formats from the network (unicast or multicast) or from disk. This also includes a generalization of the packet writer interface. This PR changes the Bifrost API in a couple of ways:

bifrost.udp_capture.UDPCapture is now bifrost.packet_capture.UDPCapture
there is a new PacketCaptureCallback class that wraps the callbacks
bifrost.udp_transmit is now bifrost.packet_writer

This should also address the segfault I was running into in #137.

PR #223, opened 2023-09-06
Update `bifrost.memory`'s understanding of memory spaces

jaycedowell:
This PR updates the "memset" and "memcpy" functions in `bifrost.memory` to work with `ndarray`s that live in non-system accessible spaces.


jaycedowell:
These test failures might be related to #220.

jaycedowell:
It looks like these failures (and some of those part of the testing for https://github.com/ledatelescope/bifrost/pull/166) are related to missing file(s) needed by the extended `testbench` tests.  Maybe we should drop those tests?

Issue #210, opened 2023-06-09
Textures being created with incorrect sizeInBytes


torrance:
The `test_linalg/test_matmul_ab_beamformer_kernel_small` test adds a `misalign` parameter to test `linalg_kernels.cu` handling of memory not aligned to 512 byte boundaries:

https://github.com/ledatelescope/bifrost/blob/30fd2682ec22bc26d9dd8cd9bc424aefc4e7075e/test/test_linalg.py#L224-L231

For example, on the third innermost iteration, this tests passes in the values `(nchan = 1, ntime = 1, nstand = 3, misalign = 2)`. This produces an array with size `(1, 1, 6)`, which is then truncated (using the misalign value) down to a `(1, 1, 4)`.

After calling `linalg.matmul`, this eventually reaches the function `bf_cherk_N()` where the total elements in the array needs to be calculated:

https://github.com/ledatelescope/bifrost/blob/30fd2682ec22bc26d9dd8cd9bc424aefc4e7075e/src/linalg_kernels.cu#L539-L540

For this particular test, the `n_element_total` should be `2 (size of truncated array) + 1 (offset)`, however it calculates it as `3 + 1 = 4`.

When this incorrect, larger size is later used to calculate the `sizeInBytes` of the texture, this results in the texture believe it has access to more memory than it really does. On CUDA backends this isn't actually validated, but on ROCM backends, the texture creation fails since it actually checks the backing memory.

I've spent some time with this but haven't been able to fix it, as I'm not clear what all of N, K, A_stride, nbatch, etc. are, or where they are meant to be coming from.

jaycedowell:
I'll try to find some time this week to look into this.

jaycedowell:
I spent some time looking into this and I *think* I know what is happening for this particular case.   I think the underlying problem is how Bifrost computes array sizes.  By default it uses the product of the first array stride length with the size of the first array dimension to get the overall array size.  In Python that would be:
```
from bifrost import ndarray
a = ndarray(shape=(1,1,6), dtype='ci8')
print(a.size*a.itemsize, 'vs', a.strides[0]*a.shape[0])
```
which looks ok.  When `misalign` is non-zero we end up with:
```
b = a[...,2:]
print(b.size*b.itemsize, 'vs', b.strides[0]*b.shape[0])
```
which doesn't match.  The `misalign=2` causes the start address for the underlying memory and the shape to be updated but not the strides.
```
[...]

PR #224, opened 2023-11-14
Update linalg algos

dentalfloss1:
Updated the linalg algorithms to tune for performance

### DIFF
--- a/src/linalg.cu
+++ b/src/linalg.cu
@@ -209,7 +209,7 @@ BFstatus bfMatMul_aa_exec(BFlinalg    handle,
 	//bool use_bf_cherk = use_bf_cherk_str && atoi(use_bf_cherk_str);
 	enum { BF_CUBLAS_CHERK_THRESHOLD = 896 };
 	if( //use_bf_cherk &&
-	    (CUDART_VERSION < 8000 || n < BF_CUBLAS_CHERK_THRESHOLD) &&
+	    n < BF_CUBLAS_CHERK_THRESHOLD &&
 	    trans == CUBLAS_OP_N &&
 	    n % 2 == 0 &&
 	    a_stride % 2 == 0 && a_batchstride % 2 == 0 &&
--- a/src/linalg.cu
+++ b/src/linalg.cu
@@ -679,15 +679,28 @@ BFstatus bfMatMul_ab_exec(BFlinalg    handle,
 	
 	// TODO: Why does ci8 yield nans when we go into batched execution?
 	if( nbatch > 12 &&  a_type != BF_DTYPE_CI8 ) {
-			BF_CHECK( bfMatMul_ab_exec_batch(handle, stream,
-																			 trans_a, trans_b,
-																			 m, n, k,
-																			 alpha,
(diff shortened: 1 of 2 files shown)


src/linalg.cu
```diff
@@ -498,16 +658,18 @@ BFstatus bfMatMul_ab_exec(BFlinalg    handle,
...
 	//char* use_bf_cgemm_str = getenv("BF_CGEMM");
 	//bool use_bf_cgemm = use_bf_cgemm_str && atoi(use_bf_cgemm_str);
+       // std::cout << "nbatch: "<<nbatch<<std::endl;
 	if( //use_bf_cgemm &&
-	    n <= 12 &&
 	    trans_a == CUBLAS_OP_T && trans_b == CUBLAS_OP_N &&
 	    (a_type == BF_DTYPE_CI4  || a_type == BF_DTYPE_CI8) &&
 	    (b_type == BF_DTYPE_CI16 || b_type == BF_DTYPE_CF16 || b_type == BF_DTYPE_CF32) &&
 	    c_type == BF_DTYPE_CF32 ) {
+                std::cout<<"custom kernel"<<endl;
```
jaycedowell, line 672: Looks like there are few debugging-style comments/print statements here.

src/linalg.cu
```diff
@@ -518,19 +680,33 @@ BFstatus bfMatMul_ab_exec(BFlinalg    handle,
...
+		  //std::cout << "nbatch: " << nbatch << std::endl;
+			BF_CHECK( bfMatMul_ab_exec_batch(handle, stream,
+																			 trans_a, trans_b,
+																			 m, n, k,
+																			 alpha,
+																			 a_data, a_type, a_stride, a_batchstride,
+																			 b_data, b_type, b_stride, b_batchstride,
+																			 beta,
+																			 c_data, c_type, c_stride, c_batchstride,
+																			 nbatch) );
```
jaycedowell, line 694: What's up with the formatting here?

src/linalg.cu
dentalfloss1: whoops. Should be fixed

src/linalg.cu
dentalfloss1: fixed

dentalfloss1:
> Should we try to do the same testing/cleanups/updates for the "aa" functions?
We can if you like. I only targeted the parts that were used by the beamformer so I wasn't sure if it was worth it

jaycedowell:
It's worth a look at what's in cuBLAS to see if there is a straight forward way to add a aa_batched function.

jaycedowell:
Slightly related - can you add the texture size change from https://github.com/ledatelescope/bifrost/issues/210 to linalg?  Otherwise I think this PR is good since it doesn't look like there is much we can do with aa_batched.

dentalfloss1:
I think it was just that one liner from Torrance right? I added it into linalg_kernels.cu Is there anything else we want to do here? If it's ready to go, I'll let someone else be the one to close/commit etc

jaycedowell:
As long as this last round of tests don't explode I think it's ready to go.

PR #225, opened 2024-01-26
Add simple packet

dentalfloss1:
Adds a simple packet format with a header composed of just a sequence number and the data being 4094 shorts.


src/formats/simple.hpp
```diff
@@ -0,0 +1,157 @@
...
+           
+           _mm_store_si128(reinterpret_cast<__m128i*>(ddst),
+					                 _mm_loadu_si128(reinterpret_cast<const __m128i*>(dsrc)));
+           _mm_store_si128(reinterpret_cast<__m128i*>(ddst+1),
+					                 _mm_loadu_si128(reinterpret_cast<const __m128i*>(dsrc+1)));
+#else
+						::memcpy(&out[pkt->src + pkt->nsrc*chan],
+						      	 &in[chan], sizeof(otype));
+#endif
+#endif
```
jaycedowell, line 112: Since we are setting `nchan` to 1 in `SIMPLEDecoder::operator()` this loop would only copy the first 32B out of a packet and into the ring buffer?

jaycedowell:
Two additional thoughts:
 1. If the optimization/intrinsic stuff isn't working it could be because it isn't implemented correctly.
 2. We could consider some kind of loopback-style test of the packet format similar to what is in `test.test_disk_io` and `test.test_udp_io`.

dentalfloss1:
The optimizations worked when I switched machines. It was probably due to issues with the machine I was using. I'll work on writing tests

jaycedowell:
The tests pass.  Maybe it's ready to merge.

PR #233, opened 2024-04-25
Update docs

dentalfloss1:
Clarify build instructions in the README file for users who are not using Docker. 

Add information on making new packet formats in the C++ development section of the docs.


jaycedowell:
Some other things to think about:
 * Should the Makefile in docs also run doxygen as part of building the docs?
 * Should there be some toggle in `configure`/a Makefile that makes things work even without doxygen?
 * Should we also test for sphinx/breathe or something like `extras_require` in the setup.py?
 * Should the top-level Makefile have a `make docs` option that builds everything if Python/doxygen are installed?

jaycedowell:
I've made some changes to `configure`.  It now:
 * emits a warning if `doxygen` isn't found about how the docs cannot be built
 * tests for the `sphinx-build`, `sphinx-apidoc`, and `breathe-apidoc` scripts
 * emits a warning if `sphinx-*` or `breathe-apidoc` are not found about how the _Python_ docs cannot be built
 * updates `make doc` to build all possible docs with the Python stuff built as "singlehtml"
 * updates docs/Makefile to a .in file so that it is consistent with all other build files and uses the configure-detected values for the paths.

I think that covers it.  We should also update master once we're happy with these doc build changes.

dentalfloss1:
I made quite a few changes to the README. Let me know what you think 
\

PR #234, opened 2024-05-19
Lockfile will warn periodically if waiting more than 5s

### LINKED ISSUE #232: Improve debuggability in case of a stale LockFile
As noted in the comment on the [LockFile constructor](https://github.com/ledatelescope/bifrost/blob/master/src/fileutils.cpp#L190), it quietly busy-waits if a stale lock file is left behind after abnormal exit.

Should be relatively easy just to emit a message after a few seconds of busy waiting (Greg's suggestion).

In addition, if we store the PID (+hostname?) as suggested in the comment, then there is more information available for troubleshooting.

Issue pointed out as part of [this ng-dp issue](https://github.com/lwa-project/ng_digital_processor/issues/30#issuecomment-1987045798).

league:
Fixes #232.

I think this is ready to go. There's a self-contained test program at the bottom of the file, it's easy to compile and test in isolation.

Is a `cerr` message good enough?

### DIFF
--- a/src/fileutils.cpp
+++ b/src/fileutils.cpp
@@ -28,6 +28,7 @@
 
 #include "fileutils.hpp"
 #include <sstream>
+#include <iostream>
 
 #if defined(HAVE_CXX_FILESYSTEM) && HAVE_CXX_FILESYSTEM
 #include <filesystem>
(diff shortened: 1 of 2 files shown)


league:
I will add the PID check to this PR.

jaycedowell:
Works for me on Ubuntu 20.04 and MacOS 12.7.

jaycedowell:
I was about to merge this when I noticed that the self-hosted builds were hanging.  Why's that happening?

jaycedowell:
Self hosted is back.

PR #236, opened 2024-05-29
Fix DeprecationWarning: invalid escape sequence \d

league:
Two places where regexes used `\d` without being "raw" strings. The one in `like_pmap` was showing up in tests as:

```
like_pmap:95: DeprecationWarning: invalid escape sequence '\d'
```

~Though it's not yet clear to me how/why `like_pmap` is being imported into any tests...?~

### DIFF
--- a/test/test_scripts.py
+++ b/test/test_scripts.py
@@ -53,7 +53,7 @@ except ImportError:
     pass
 run_scripts_tests &= (sys.version_info[0] >= 3)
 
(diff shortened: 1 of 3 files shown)


league:
For reference, `like_pmap.py` is being tested from `test_scripts.py`, with a glob over `TOOLS_DIR`.

jaycedowell:
Yes, it's test_scripts.py that looks at those files.  It's an attempt to use static analysis with pylint to find problems with  the scripts without having to run them.

Issue #201, opened 2023-04-17
CUDA 11.8 + bifrost.fir = error


jaycedowell:
I was testing under CUDA 11.8 and I kept running into a:
```
fir.cu:248 cudaGetErrorString(cuda_ret) = too many resources requested for launch
fir.cu:248 Condition failed: cuda_ret == cudaSuccess
fir.cu:248 error 99: BF_STATUS_INTERNAL_ERROR
fir.cu:413 error 99: BF_STATUS_INTERNAL_ERROR
```
This error disappears under CUDA 11.6 and everything acts the way it should.

telegraphic:
Hey Jayce,

Could be related this this tale from Matt: _"Long story short, this was failing in the hashpipe rawspec thread and the solution was we had to downgrade from cuda 11.8 to 11.7 - thanks to Luigi (who happened to also be traveling here at GB) for pointing out that 11.8 has bugs that would cause CUFFT failures. Once Dave downgraded, we were recording data again."_

telegraphic:
Not sure what "bugs" are however, but CUDA 11.8 is forsaken...

jaycedowell:
Since we are working on CUDA 12 support trying to fight with 11.8 isn't worth it.  If this pops up again then maybe we add a catch to `configure` to throw an error if 11.8 is found.

PR #244, opened 2024-07-18
Add cudaver to docs

dentalfloss1:
Adds a link to the docs where a table of test results of the CUDA toolkit live. The table has been added to the docs, but of course, we'll also need to update the hosted version of the docs.


jaycedowell:
Two questions:
 * What does "likely working" mean?
 * The comments in #201 imply that CUDA 11.8 doesn't actually work but I also see this listed as likely working in the table.  Does the test suite pass?

dentalfloss1:
Perhaps we can just change "likely working" to "working." The card I was using sometimes puts values just out of spec for the tests, so technically no tests would fully pass. It's weird and confusing. Let's just upgrade all "likely working" to "working." 
That bug is interesting. All the tests passed I believe on 11.8. But, if we have an existing issue with 11.8 perhaps we should just say "not working" to be safe. 
What do you think?

jaycedowell:
Yes and yes.  I'll update the table.

jaycedowell:
What do you think about dropping the "CUDA_ERROR_UNSUPPORTED_PTX_VERSION" entries?  From what I've read those come from mis-matched kernel driver and toolkit versions (one is too old for the other).  It's not really a CUDA version thing or a Bifrost thing but more of an installation thing.

dentalfloss1:
Sounds good. Maybe we can just add a footnote that errors can occur from mismatched drivers?

Sent from my iPhone

On Jul 22, 2024, at 1:12 PM, jaycedowell ***@***.***> wrote:

﻿
  [EXTERNAL]

What do you think about dropping the "CUDA_ERROR_UNSUPPORTED_PTX_VERSION" entries? From what I've read those come from mis-matched kernel driver and toolkit versions (one is too old for the other). It's not really a CUDA version thing or a Bifrost thing but more of an installation thing.

PR #250, opened 2024-11-11
Switch to universal-ctags

### LINKED ISSUE #246: Bifrost requires exuberant-ctags package. Not available on Rocky 9.4
Hi - I tried following the instructions of installing bifrost on a Rocky 9.4 Linux VM and came up against an issue. The installation requires the use of a package called exuberant-ctags and it is not available to Rocky 9.4. It is however part of the base installation of Ubuntu OS. 
This seems to suggest that bifrost will not be compatible with Rocky Linux. I came up against this issue a few weeks ago but should do another test install just to verify what I'm writing here is 100% correct.
JJ

jaycedowell:
Fixes #246 at some level.

This PR moves over to using universal-ctags.  This seems to be a successor to exuberant-ctags that is available on Ubuntu, home brew, and Redhat-based system.

Things left to do:
- [x] Update `configure` so that it isn't explicitly tied to exuberant-ctags
- [x] Update the documentation
- [x] Update `flake.nix`
- [ ] Update the Dockerfiles (also see #215)
- [x] Make sure the tutorial still runs on Google Colab

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
             build-essential \
             ca-certificates \
             curl \
-            exuberant-ctags \
+            universal-ctags \
             gfortran \
             git \
             libhwloc-dev \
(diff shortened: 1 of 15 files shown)


jaycedowell:
@league where do things stand with #215?  I see that it is still a draft but how drafty of a draft is it?

jaycedowell:
The "getting started" tutorial seems to work on Colab.

jaycedowell:
I don't see anything in `flake.nix` that specifies which kind of ctags to use.  Maybe that's a non issue.

jaycedowell:
I'm also going to punt on the dockerfiles thing.  We can address that in #215.

